PR #21, opened 2023-10-03
tidy up dependencies

jangorecki:
PR includes following changes:

- move DBI from Depends to Imports in DCF:
It is good practice to avoid Depends when possible in favor of Imports. Side effect is that we need to explicitly load DBI namespace if we want to use functions from it in examples
- utils removed from DCF:
No need to specify it. It is enough if it is in NAMESPACE
- removed importFrom DBI in NAMESPACE:
DBI was imported completely already in NAMESPACE so this line is redundant
- test script correctly escapes suggested package:
as per recommendation in [WRE](https://cran.r-project.org/doc/manuals/r-release/R-exts.html) "...recommendation to use suggested packages conditionally in tests does also apply to packages used to manage test suites: a notorious example was testthat..."

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -43,11 +43,10 @@ License: MIT + file LICENSE
 URL: https://duckdb.org/, https://github.com/duckdb/duckdb-r
 BugReports: https://github.com/duckdb/duckdb-r/issues
 Depends:
-    DBI,
     R (>= 3.6.0)
 Imports:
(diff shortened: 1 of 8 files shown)


krlmlr:
Thanks. Good catch on the `@importFrom`, I'll cherry-pick that. 

The other points are valid, but also not very relevant. Shelving them for now.

PR #35, opened 2023-10-18
Removing minus one that compiler complained about

hannes:


hannes:
Its now slightly over allocating but its not relevant

hannes:
@krlmlr this should fix part two

krlmlr:
Fixed upstream, let's import that one commit here.

hannes:
can you do that please?

PR #40, opened 2023-11-06
feat: Add `prod()` translation for dbplyr

m-muecke:


krlmlr:
Thanks. What would a test for this and other functionality look like?

m-muecke:
> Thanks. What would a test for this and other functionality look like?

I have added some tests for the current agg functions. What is the scope difference between the duckdb and the duckplyr package for `dplyr` translations, both seem to have overlap.

krlmlr:
duckplyr is data frame in, data frame out, with full compatibility and fallback to dplyr if needed.

krlmlr:
Thanks!

PR #49, opened 2023-12-03
fix: Align the argument order of `dbWriteTable()` with the DBI specs

### LINKED ISSUE #43: dbWriteTable's arguments order
The DBI document says:

> The following arguments are not part of the dbWriteTable() generic (to improve compatibility across backends) but are part of the DBI specification:
>
> - row.names (default: FALSE)
> - overwrite (default: FALSE)
> - append (default: FALSE)
> - field.types (default: NULL)
> - temporary (default: FALSE)
>
>They must be provided as named arguments. See the "Specification" and "Value" sections for details on their usage.

But now `dbWriteTable__duckdb_connection_character_data.frame` can specify these options without names.
I think we need to change the position of the dot dot dot `...`.

https://github.com/duckdb/duckdb-r/blob/8386b9cd53ad08660e00aa4f755ea7032ca25dff/R/dbWriteTable__duckdb_connection_character_data.frame.R#L9-L17

eitsupi:
Fix #43

### DIFF
--- a/R/dbWriteTable__duckdb_connection_character_data.frame.R
+++ b/R/dbWriteTable__duckdb_connection_character_data.frame.R
@@ -9,12 +9,12 @@
 dbWriteTable__duckdb_connection_character_data.frame <- function(conn,
                                                                  name,
                                                                  value,
+                                                                 ...,
                                                                  row.names = FALSE,
                                                                  overwrite = FALSE,
                                                                  append = FALSE,
                                                                  field.types = NULL,
-                                                                 temporary = FALSE,
-                                                                 ...) {
+                                                                 temporary = FALSE) {
   check_flag(overwrite)
   check_flag(append)
   check_flag(temporary)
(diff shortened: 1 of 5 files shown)


eitsupi:
@krlmlr Is there any reason why this would not be merged? Is there anything I can do?

krlmlr:
Thanks. Running revdepchecks now.

krlmlr:
Uploaded results, one package has new failures. Could you please take a look?

eitsupi:
> Uploaded results, one package has new failures. Could you please take a look?

I have looked and found that the results were also different in duckdb installed from R-universe, so I think this is due to a change in the order of the records returned by duckdb and this PR is irrelevant.
https://github.com/darwin-eu-dev/PatientProfiles/blob/97f8a26ffede8ce09c98aece9ca67c43fa97db28/tests/testthat/test-addCategories.R#L25-L30

eitsupi:
I've send a PR for fix the test cases darwin-eu-dev/PatientProfiles#456.

krlmlr:
Thanks, also for double-checking!

PR #54, opened 2023-12-11
Fix set operation tests

Tmonster:
PR to fix the set operation tests that were broken by https://github.com/duckdb/duckdb/pull/9636

This defaults the `setop_all` to true for all set operations. If users want the results to be distinct, they will have to push the distinct value on top themselves.

Also make some of the tests more robust by adding order by clauses.


krlmlr:
Thanks! I'll cherry-pick this to where the breaking occurs during the code update.

krlmlr:
Now part of #53.

PR #68, opened 2024-02-05
Tweaks for building on Windows/aarch64 with LLVM 17.

kalibera:
__umulh is a Microsoft intrinsic not supported by clang (LLVM17) with mingw target. This patch makes it used only with Visual studio.

thrift-config.h had to be updated to assert that the platform has an arithmetic right shift - on Windows (unlike other platforms), it asserted that only with x86/x86_64. I have ran the upstream thrift configure test, which concluded that the right shift is arithmetic on Windows/aarch64.

With the patch applied, the (CRAN version of the) package builds fine on my Windows/aarch64.

### DIFF
--- a/src/duckdb/third_party/fast_float/fast_float/fast_float.h
+++ b/src/duckdb/third_party/fast_float/fast_float/fast_float.h
@@ -242,7 +242,7 @@ fastfloat_really_inline uint64_t _umul128(uint64_t ab, uint64_t cd,
 fastfloat_really_inline value128 full_multiplication(uint64_t a,
                                                      uint64_t b) {
   value128 answer;
-#ifdef _M_ARM64
+#if defined(FASTFLOAT_VISUAL_STUDIO) && defined(_M_ARM64)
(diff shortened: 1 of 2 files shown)


Tmonster:
Hi Tomas,

Thank you for the PR! This PR seems to target files in DuckDB source. For this repo, all files in the `src/` directory are actually just a mirror of the duckdb/duckdb repository. (why we don't use submodules is a long story). Would you mind opening the same PR there? The changes will eventually show up in duckdb-r as well

Thanks!

kalibera:
> Hi Tomas,
> 
> Thank you for the PR! This PR seems to target files in DuckDB source. For this repo, all files in the `src/` directory are actually just a mirror of the duckdb/duckdb repository. (why we don't use submodules is a long story). Would you mind opening the same PR there? The changes will eventually show up in duckdb-r as well

Thanks, it has been already reported upstream by Hannes (https://github.com/duckdb/duckdb/pull/10479)

krlmlr:
Thanks! Will be picked up when we update the bundled sources.

PR #73, opened 2024-02-13
chore: Change directory location for extensions and secrets for v.0.10.0 release

Tmonster:


Tmonster:
Smoke test is failing because source code for v0.10.0 has not yet been merged

krlmlr (commented):
Thanks. Do we currently have a way to fetch the version of the bundled duckdb sources?

krlmlr:
Yes:

``` r
con <- DBI::dbConnect(duckdb::duckdb(), ":memory:")
DBI::dbGetQuery(con, "PRAGMA version")
#>   library_version  source_id
#> 1  v0.9.3-dev3731 d4c774b1f1
```

<sup>Created on 2024-02-24 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Can you please help me understand the nature of the config vs. the pragmas. Is the config private to the database connection, or shared across connections initiated with the same `duckdb()` driver?

krlmlr:
We should embed the version of the bundled code in the R code too so that it is available without asking duckdb.

Tmonster:
> Can you please help me understand the nature of the config vs. the pragmas. Is the config private to the database connection, or shared across connections initiated with the same `duckdb()` driver?

The config is private to the database connection. A config then stores a `DBConfigOptions` which then stores certain settings set by `pragma` function calls. Each config also stores information about extensions. If multiple read connections are created, each one could have different `DBConfigOptions` or extensions, which is why each config is private to the database.

Tmonster:
@krlmlr you might need to run the tests again. I'm trying to push empty commits to force the CI to run again, but GitHub doesn't seem to be picking it up

tests/testthat/test_extension_path.R
```diff
@@ -0,0 +1,8 @@
+skip_on_cran()
+skip_on_os(c("windows"))
+
+test_that("duckdb extensions are in a R specific directory", {
+  con <- DBI::dbConnect(duckdb(config=list('allow_unsigned_extensions'='true')))
+  on.exit(DBI::dbDisconnect(con, shutdown = TRUE))
+  expect_error(dbExecute(con, "LOAD 'httpfs'"),regexp = ".*R/duckdb/extensions/.*/httpfs.duckdb_extension\" not found.")
```
carlopi, line 7: Can this test be possibly done on a non-existing extension? Something like `LOAD 'not_existing_extension'` and checking for `duckdb/extensions/.*/not_existing_exension.duckdb_extension` in the error message?
Unsure in what conditions this test is run, but I think this will fail if httpfs happens to be there.

krlmlr:
Thanks!

PR #86, opened 2024-02-24
NOT FOR MERGING: Isolate hang in Arrow test

krlmlr:
This PR shows a minimal change that leads to `dbSendQuery(arrow = TRUE)` hanging.

Test code:

```r
con <- DBI::dbConnect(duckdb::duckdb(), ":memory:")
DBI::dbSendQuery(con, "SELECT 1", arrow = TRUE)
```

On the main branch of this repo, this code works. On the commit "Harmless parts" from this PR, the code works. On the commit "Remove special case", it does not terminate.

This change is part of duckdb v0.10.0. I'd like to resolve this before proceeding with the CRAN release.

I also tested with `SET threads = 1`, same result.

Upstream: https://github.com/duckdb/duckdb/pull/10245#issuecomment-1961759356

The first commit in this PR corresponds to the main branch just before merging https://github.com/duckdb/duckdb/pull/10245. The last two commits are a decomposition of https://github.com/duckdb/duckdb/commit/68c9e1afaf8fab255fbd94c5c0144ee2f1f20fa7.

@Tmonster: Can you replicate

PR #89, opened 2024-02-26
Fix Hanging query

Tmonster:
As mentioned in https://github.com/duckdb/duckdb-r/pull/86, https://github.com/duckdb/duckdb/pull/10245 would cause a simple case in duckdb-r to hang. This PR presents a solution that better addresses the infinite loop issue. Apparently Query results can be `blocked` or `finished` if `allow_streaming_result = true`. For arrow, we pass `true` for streaming result, so we need to check is the result is blocked or finished instead of just finished. A blocked query just means `fetch` needs to be called.


krlmlr:
Thanks. Will 010d6af26e187ca45ff204a060f907d43ad4d70f on top of #86 fix the issue? I'll investigate.

Tmonster:
> Will [010d6af](https://github.com/duckdb/duckdb-r/commit/010d6af26e187ca45ff204a060f907d43ad4d70f) on top of #86 fix the issue?

Yes that would fix the issue

Are https://github.com/duckdb/duckdb-r/pull/86/commits/efc7cd54a757d1b6b8bbd87977cd14f753519d0c and https://github.com/duckdb/duckdb-r/pull/86/commits/ec2cf28c23e48b18114ae60e600eaa04473d318b necessary? The fix should work just by itself as well (although I did not do a full regression test, so I'm not sure)

krlmlr:
Done in #90. Thanks!

PR #90, opened 2024-02-27
feat: Update core to duckdb v0.10.0

krlmlr:
With all tests passing.


src/statement.cpp
```diff
@@ -337,7 +337,7 @@ bool FetchArrowChunk(ChunkScanState &scan_state, ClientProperties options, Appen
 	do {
 		execution_result = pending_query->ExecuteTask();
 		R_CheckUserInterrupt();
-	} while (!PendingQueryResult::IsFinished(execution_result));
+	} while (!PendingQueryResult::IsFinishedOrBlocked(execution_result));
```
krlmlr, line 340: @Tmonster: I cherry-picked this for #90, this seemed necessary due to some change in the core. This now breaks again, #190, after duckdb/duckdb#12636. Is this safe to revert, what do you think?

CC @Tishj.

PR #103, opened 2024-03-12
Remove std:: from unordered_map

lnkuiper:
We're using namespace `duckdb` which uses it so it should be fine. I plan to make changes to `duckdb::unordered` map in a separate PR to the duckdb repo. I hope this fixes my CI!


krlmlr (commented):
Thanks!

PR #108, opened 2024-03-13
Fix invalid regex in .Rbuildignore

Tmonster:
An invalid regex in the .RbuildIgnore is causing many CI failures. A `*` regex character cannot match the `^` character 0 or more times. 
I've changed it to `.*` but I'm not sure if the original intention is to match everything with the suffix `_files`


Mytherin:
Thanks!

krlmlr (commented):
Thanks, good catch!

PR #122, opened 2024-03-23
feat: Implement `n_distinct()` for multiple arguments using duckdb structs

### LINKED ISSUE #110: Feature request: `n_distinct()` implementation w/ multiple arguments.
Hi,

I understand that `n_distinct()` with multiple arguments is not supported for a generic dbplyr backend since SQL does not support that out of the box.

Is it possible to implement it in the case of duckdb using [structs](https://duckdb.org/docs/sql/data_types/struct)?

For example:
``` r
dbplyr::tbl_lazy(data.frame(x = 1, y = 1), duckdb::simulate_duckdb()) |>
summarize(l = n_distinct(x, y))
```
could be translated to
``` SQL
<SQL>
SELECT COUNT(DISTINCT {'x' : `x`, 'y' : `y`}) AS `l`
FROM `df`
```
which executes on duckdb.

lschneiderbauer:
This PR attempts to adress #110, implementing `n_distinct()` making use of duckdb's [structs](https://duckdb.org/docs/sql/data_types/struct).

I don't have any experience in writing dbplyr SQL translations, in particular I have no idea of the 'best practices', but I tried to mimic as much from the already existing code as I could.

A couple of notes:

- duckdb's struct fields need to be named. In this case, the naming is irrelevant and I chose to name them 'v1', 'v2', 'v3', ... and so on, depending on the number of arguments passed to `n_distinct()`. If there is an alternative prefered way to do it, please let me know.
- I was making use of `glue::glue()`. Although dbplyr depends on glue, and in this function we are assuming that dbplyr is installed, I am not sure if I am allowed to assume that this function exists. Maybe I should not make use of it?
- I also added unit tests at places where I thought they make sense. Let me know if I should change them or relocate them, etc..

Thanks! :)

Closes #110

### DIFF
--- a/R/backend-dbplyr__duckdb_connection.R
+++ b/R/backend-dbplyr__duckdb_connection.R
@@ -80,7 +80,7 @@ duckdb_grepl <- function(pattern, x, ignore.case = FALSE, perl = FALSE, fixed = 
 duckdb_n_distinct <- function(..., na.rm = FALSE) {
   sql <- pkg_method("sql", "dbplyr")
 
-  if (identical(na.rm, TRUE)) {
+  if (!identical(na.rm, FALSE)) {
     stop("Parameter `na.rm = TRUE` in n_distinct() is currently not supported in DuckDB backend.", call. = FALSE)
   }
 
--- a/R/backend-dbplyr__duckdb_connection.R
+++ b/R/backend-dbplyr__duckdb_connection.R
@@ -77,6 +77,26 @@ duckdb_grepl <- function(pattern, x, ignore.case = FALSE, perl = FALSE, fixed = 
   }
 }
 
+duckdb_n_distinct <- function(..., na.rm = FALSE) {
+  check_na_rm <- pkg_method("check_na_rm", "dbplyr")
+  glue_sql2 <- pkg_method("glue_sql2", "dbplyr")
+
+  check_na_rm(na.rm)
+
+  vars <- list(...)
+  str_struct <-
+    paste0("{", paste0(
+      lapply(
+        seq_along(vars),
+        \(i) glue::glue("'v{i}' : {vars[[i]]}")
+      ),
+      collapse = ", "
+    ), "}")
+  glue_sql2(
+    sql_current_con(),
+    "COUNT(DISTINCT {str_struct})"
+  )
+}
 
 # Customized translation functions for DuckDB SQL
 # @param con A \code{\link{dbConnect}} object, as returned by \code{dbConnect()}
@@ -316,7 +336,8 @@ sql_translation.duckdb_connection <- function(con) {
       any = sql_aggregate("BOOL_OR", "any"),
       str_flatten = function(x, collapse) sql_expr(STRING_AGG(!!x, !!collapse)),
       first = sql_prefix("FIRST", 1),
-      last = sql_prefix("LAST", 1)
+      last = sql_prefix("LAST", 1),
+      n_distinct = duckdb_n_distinct
     ),
     sql_translator(
       .parent = base_win,
@@ -333,7 +354,14 @@ sql_translation.duckdb_connection <- function(con) {
           partition = win_current_group(),
           order = win_current_order()
         )
-      }
+      },
+      n_distinct =
+        function(..., na.rm = FALSE) {
+          win_over(
+            duckdb_n_distinct(..., na.rm = na.rm),
+            partition = win_current_group()
+          )
+        }
     )
   )
 }
--- a/R/backend-dbplyr__duckdb_connection.R
+++ b/R/backend-dbplyr__duckdb_connection.R
@@ -78,25 +78,16 @@ duckdb_grepl <- function(pattern, x, ignore.case = FALSE, perl = FALSE, fixed = 
 }
 
 duckdb_n_distinct <- function(..., na.rm = FALSE) {
-  glue_sql2 <- pkg_method("glue_sql2", "dbplyr")
+  sql <- pkg_method("sql", "dbplyr")
 
   if (identical(na.rm, TRUE)) {
     cli::cli_abort("`na.rm = TRUE` not implemented.")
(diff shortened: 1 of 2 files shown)


R/backend-dbplyr__duckdb_connection.R
```diff
@@ -77,6 +77,26 @@ duckdb_grepl <- function(pattern, x, ignore.case = FALSE, perl = FALSE, fixed =
   }
 }
 
+duckdb_n_distinct <- function(..., na.rm = FALSE) {
+  check_na_rm <- pkg_method("check_na_rm", "dbplyr")
+  glue_sql2 <- pkg_method("glue_sql2", "dbplyr")
```
krlmlr, line 82: Those are private functions. @mgirlich: Is it fair to assume that these will stay around?

krlmlr (commented):
Thanks! A small integration test (what does `n_distinct()` actually do with a real dummy table) would be useful too.

Sorry about the conflicts.

R/backend-dbplyr__duckdb_connection.R
mgirlich: `check_na_rm()` should definitely stay and might be worth exporting for backends. I opened an issue: https://github.com/tidyverse/dbplyr/issues/1483.
It would also make sense to export `glue_sql2()` but I don' think this is ready yet. So, I guess this is more likely to evolve (in particular I would consider changing the name before exporting it).

lschneiderbauer:
> A small integration test (what does `n_distinct()` actually do with a real dummy table) would be useful too.
I added a bunch of tests to check the actual computation.

In the meantime I also realized that my assumption that duckdb would use a `na.rm = TRUE` semantics in this case is actually wrong: NULLs within structs seem to actually be counted as a single unique entity. The added test results suggest that.
So I removed the wrong `check_na_rm()` and opted to throw an error if the `na.rm = TRUE` is passed. Since `na.rm = FALSE` is the default this seemed appropriate.
If one wanted to implemented a `na.rm = TRUE` behavior, using the [`FILTER`](https://duckdb.org/docs/sql/query_syntax/filter) clause might be a way to go, but according to the documentation this does not work within a windowing context. Also, it's not obvious to me how to check any element in a struct for NULLs with the [struct functions](https://duckdb.org/docs/sql/functions/nested.html#struct-functions) that are available so far.

Regarding the use of `dbplyr:::glue_sql2()`, what should be the way to go: Wait with the PR until such a function becomes exported, reimplement it in this package, or keep using the private function?

krlmlr:
Thanks. If we must, we can reimplement here, or use plain `paste0()` ?

lschneiderbauer:
Thanks.

- I am using the `dbplyr::sql()` function directly now, together with `paste0()`, instead of `dbplyr:::glue_sql2()`.
- `cli::cli_abort()` is replaced by `stop()` to get rid of R CMD CHECK warning.
- I am making use of duckdb's [row()](https://duckdb.org/docs/sql/data_types/struct.html#creating-structs-with-the-row-function) function now instead of using the '{ ... }' syntax which makes generating SQL easier. (and also gets rid of `glue::glue()` to make R CMD CHECK happy)

R/backend-dbplyr__duckdb_connection.R
```diff
@@ -77,6 +77,18 @@ duckdb_grepl <- function(pattern, x, ignore.case = FALSE, perl = FALSE, fixed =
   }
 }
 
+duckdb_n_distinct <- function(..., na.rm = FALSE) {
+  sql <- pkg_method("sql", "dbplyr")
+
+  if (identical(na.rm, TRUE)) {
```
krlmlr, line 83: Slightly stricter:

```suggestion
  if (!identical(na.rm, FALSE)) {
```

lschneiderbauer:
@krlmlr Anything else you would like me to improve?

R/backend-dbplyr__duckdb_connection.R
```diff
@@ -77,6 +77,18 @@ duckdb_grepl <- function(pattern, x, ignore.case = FALSE, perl = FALSE, fixed =
...
+  sql <- pkg_method("sql", "dbplyr")
+
+  if (!identical(na.rm, FALSE)) {
+    stop("Parameter `na.rm = TRUE` in n_distinct() is currently not supported in DuckDB backend.", call. = FALSE)
+  }
+
+  # https://duckdb.org/docs/sql/data_types/struct.html#creating-structs-with-the-row-function
+  str_struct <- paste0("row(", paste0(list(...), collapse = ", "), ")")
+
+  sql(paste0("COUNT(DISTINCT ", str_struct, ")"))
```
krlmlr, line 90: @mgirlich: Can you please help me review this part of the code?

krlmlr (commented):
Thanks for the nudge!

tests/testthat/test-backend-dbplyr__duckdb_connection.R
```diff
@@ -218,8 +221,54 @@ test_that("two variable aggregates are translated correctly", {
...
+    pull(summarize(df, n = n_distinct(x, y)), n),
+    3
+  )
+
+  # two columns containing NAs return correct results
+  expect_equal(
+    pull(summarize(df_na, n = n_distinct(x, y)), n),
+    5
+  )
+})
```
mgirlich, line 271: You might want to add a test that uses a window.

R/backend-dbplyr__duckdb_connection.R
mgirlich: Basically looks good to me. This also works for diverse user input, e.g. `n_distinct(x, a / (b - 1))` or `n_distinct(!!x)`.
One minor thing: you might want to check for empty dots.

mgirlich (approved):

krlmlr:
Thanks! Adding the extra tests and checks could be done in a small follow-up PR.

PR #138, opened 2024-04-03
fix: `rel_sql(rel, "{{sql}}")` works even on a read-only database

Tmonster:
If a user attempts to query a read only database with `rel_sql` they will encounter an error. This is because `rel_sql` creates a view with name `_`. 

The fix is to make this a temporary view. If the user still tries to create a table with `create table as (select * from _)` an error should still be thrown

Fixes https://github.com/duckdblabs/duckdb-internal/issues/1721


krlmlr:
Thanks!

src/duckdb/src/main/relation.cpp
```diff
@@ -324,7 +324,9 @@ unique_ptr<QueryResult> Relation::Query(const string &sql) {
 }
 
 unique_ptr<QueryResult> Relation::Query(const string &name, const string &sql) {
-	CreateView(name);
+    bool replace = true;
+    bool temp = IsReadOnly();
+	CreateView(name, replace, temp);
```
krlmlr, line 329: @Tmonster: I realized too late that this patches the duckdb sources. Has this been upstreamed by now? Do you want to?

I can work around with a patch file for now.

Tmonster: has not been upstreamed yet, and might not be until after v1.0 comes out. I think a patch file in duckdb-r is the way to go for now. I will still create an upstream PR though so you can eventually remove that patch.

PR #143, opened 2024-04-24
feat: Map `NA` to `SQLNULL`

krlmlr:
Is this a reasonable change? This helps with at least one package with duckplyr.

### DIFF
--- a/src/include/typesr.hpp
+++ b/src/include/typesr.hpp
@@ -115,7 +115,7 @@ struct RApiTypes {
 	static LogicalType LogicalTypeFromRType(const RType &rtype, bool experimental);
 	static string DetectLogicalType(const LogicalType &stype, const char *caller);
 	static R_len_t GetVecSize(RType rtype, SEXP coldata);
-	static Value SexpToValue(SEXP valsexp, R_len_t idx);
+	static Value SexpToValue(SEXP valsexp, R_len_t idx, bool typed_logical_null = true);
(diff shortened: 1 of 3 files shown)


krlmlr:
As it stands, this is a breaking change:

## v0.10.1

``` r
library(duckdb)
#> Loading required package: DBI
con <- dbConnect(duckdb::duckdb())

out <- dbGetQuery(con, "SELECT ? AS a", params = list(NA))
dput(out$a)
#> NA
```

<sup>Created on 2024-04-24 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

## v0.10.2 RC

``` r
...
dput(out$a)
#> NA
```

<sup>Created on 2024-04-24 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

## This PR

``` r
...
dput(out$a)
#> NA_integer_
```

Any other failure modes you could think of, Hannes?

krlmlr:
With the latest change:

``` r
...
dput(out$a)
#> NA
```

<sup>Created on 2024-04-24 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Still wondering if `SQLNULL` is the way to go here. These are the semantics of dplyr, anyway: whenever we see a single `NA`, it is treated as an untyped missing value.

krlmlr:
Need to see about the R-devel checks. Merging now, review still appreciated.

PR #144, opened 2024-04-26
feat: Include rfuns extension

krlmlr:


krlmlr:
@samansmink: I'd like to statically link the rfuns extension with the duckdb R package. Do you see downsides in the way this PR solves it?

- Expose an entry point for the R code to register the linked extension
- Do not list the extension as part of the built-in extensions

The alternative I see is to just register the functions this extension provides, without registering the extension itself.

samansmink (commented):
Hey @krlmlr!

First of all; I haven't really looked in to duckdb-r nor the rfunctions extension, so i'm not super familiar here.

In general I think this looks fine. Inlining like this may not be super scalable from a maintainability perspective, but if the requirement of linking an extension statically here is expected to be a one-off thing, I think its ok.

> Do not list the extension as part of the built-in extensions
Do you mean this as an intended requirement or as a consequence of the current implementation? If the goal is to have the rfuns extension show up in `from duckdb_extensions()` I think having it marked as a statically linked extension would be the cleanest. Note that this PR https://github.com/duckdb/duckdb/pull/11677 should ensure it will be correctly marked as a statically linked extension 

I think in general to inline the rfuns in duckdb-r it is probably cleanest do this as an extension as you have done now: In the future we may improve duckdb's internal documention functionality and then it would be nice to be able to communicate to users that these functions come from an extension called `rfuns` that is statically loaded.

samansmink:
@krlmlr installing is fine, loading should be a NOP if the extension is loaded in the way you do in this PR where it would get added to `DatabaseInstance::loaded_extensions`

PR #151, opened 2024-04-30
Change conversion for difftime to INTERVAL, not TIME

hannes:
This PR changes the DuckDB representation of `difftime` R objects from `TIME` to `INTERVAL`. The reasoning here is that R's difftime can represent way larger time differences than single days, which is what `TIME` is for. This goes wrong if the difftime is longer than one day and one tries for example casting the value to a string.

Here is an example that demonstrates the problem

```R
con <- DBI::dbConnect(duckdb::duckdb())
duckdb::duckdb_register(con, "difftime", data.frame(difftime=structure(710, units = "days", class = "difftime")))
print(DBI::dbGetQuery(con, 'select difftime, difftime::varchar difftime_string from difftime'))
```

Currently, this returns random garbage for the string
```
       difftime difftime_string
1 61344000 secs  \xd2\034:00:00
``` 

and triggers an assertion when built in debug mode.

With this PR, it returns
```
       difftime difftime_string
1 61344000 secs        710 days
```

### DIFF
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -99,15 +99,15 @@ RType RApiTypes::DetectRType(SEXP v, bool integer64) {
 		}
 		SEXP units0 = STRING_ELT(units, 0);
 		if (units0 == RStrings::get().secs) {
-			return RType::TIME_SECONDS;
+			return RType::INTERVAL_SECONDS;
 		} else if (units0 == RStrings::get().mins) {
-			return RType::TIME_MINUTES;
+			return RType::INTERVAL_MINUTES;
 		} else if (units0 == RStrings::get().hours) {
-			return RType::TIME_HOURS;
+			return RType::INTERVAL_HOURS;
 		} else if (units0 == RStrings::get().days) {
-			return RType::TIME_DAYS;
+			return RType::INTERVAL_DAYS;
 		} else if (units0 == RStrings::get().weeks) {
-			return RType::TIME_WEEKS;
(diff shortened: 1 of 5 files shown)


hannes:
we've always returned `difftime` for `TIME` because AFAIK R does not have another suitable time type.

src/types.cpp
```diff
@@ -332,24 +331,24 @@ timestamp_t RTimestampType::Convert(double val) {
 	return Timestamp::FromEpochMicroSeconds(round(val * Interval::MICROS_PER_SEC));
 }
 
-dtime_t RTimeSecondsType::Convert(double val) {
-	return dtime_t(int64_t(val * Interval::MICROS_PER_SEC));
+interval_t RIntervalSecondsType::Convert(double val) {
+	return Interval::FromMicro(int64_t(val * Interval::MICROS_PER_SEC));
```
krlmlr, line 335: This is consistent with the old implementation, but when converting datetimes I think we're rounding. Should we change this as part of the change here and further below?

```suggestion
	return Interval::FromMicro(int64_t(round(val * Interval::MICROS_PER_SEC)));
```

krlmlr (approved):
Looks good, just one question.

krlmlr:
How do I construct a literal interval of 1.5 hours?

I'll try to formulate issues upstream.

krlmlr:
No more objections after reading the docs. Tests are failing locally, need to investigate.

PR #157, opened 2024-05-05
feat: `round()` duckdb translation uses `ROUND_EVEN()` instead of `ROUND()`

### LINKED ISSUE #146: `round()` semantics of R-version differs from its duckdb-translation
In the current state of duckdb-r, the `round()` semantics of R-version differs from its duckdb-translation in case where the rounded digit is "5".

See the difference in the following reprex:

``` r
library(dplyr)
#> 
#> Attache Paket: 'dplyr'
#> Die folgenden Objekte sind maskiert von 'package:stats':
#> 
#>     filter, lag
#> Die folgenden Objekte sind maskiert von 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(duckdb)
#> Lade nötiges Paket: DBI
library(DBI)


df <- tibble(x = c(0.5, 1.5))

df |>
  mutate(
    r = round(x, 0L)
  )
#> # A tibble: 2 × 2
#>       x     r
#>   <dbl> <dbl>
#> 1   0.5     0
#> 2   1.5     2

con <- DBI::dbConnect(duckdb())
duckdb::duckdb_register(con, "df", df)

tbl(con, "df")  |>
  mutate(
    r = round(x, 0L),
    r2 = round_even(x, 0L)
  )
#> # Source:   SQL [2 x 3]
#> # Database: DuckDB v0.10.0 [xxx@Windows 10 x64:R 4.3.2/:memory:]
#>       x     r    r2
#>   <dbl> <dbl> <dbl>
#> 1   0.5     1     0
#> 2   1.5     2     2

DBI::dbDisconnect(con)
```

<sup>Created on 2024-04-29 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

From the `round()` documentation in R:

> Note that for rounding off a 5, the IEC 60559 standard (see also ‘IEEE 754’) is expected to be used, ‘go to the even digit’. Therefore round(0.5) is 0 and round(-1.5) is -2. However, this is dependent on OS services and on representation error (since e.g. 0.15 is not represented exactly, the rounding rule applies to the represented number and not to the printed number, and so round(0.15, 1) could be either 0.1 or 0.2).

The standard duckdb implementation does not adhere to this standard (see e.g. the discussion in [duckdb issue](https://github.com/duckdb/duckdb/issues/2348)). But an alternative implementation does this: [`round_even()`](https://duckdb.org/docs/sql/functions/numeric.html#round_evenv-numeric-s-integer).

Should the SQL-translation translate R "round()" to "round_even()" in order to get consistent results?

lschneiderbauer:
* In addition, I adjusted the default value of the digits argument in `round()` to `round(x, digits = 0)` to be consistent with the R default value.

Fixes #146

### DIFF
--- a/R/backend-dbplyr__duckdb_connection.R
+++ b/R/backend-dbplyr__duckdb_connection.R
@@ -461,4 +461,4 @@ tbl_query <- function(src, query, ..., cache = FALSE) {
 }
 
 # Needed to suppress the R CHECK notes (due to the use of sql_expr)
-utils::globalVariables(c("REGEXP_MATCHES", "CAST", "%AS%", "INTEGER", "XOR", "%<<%", "%>>%", "LN", "LOG", "ROUND", "EXTRACT", "%FROM%", "MONTH", "STRFTIME", "QUARTER", "YEAR", "DATE_TRUNC", "DATE", "DOY", "TO_SECONDS", "BIGINT", "TO_MINUTES", "TO_HOURS", "TO_DAYS", "TO_WEEKS", "TO_MONTHS", "TO_YEARS", "STRPOS", "NOT", "REGEXP_REPLACE", "TRIM", "LPAD", "RPAD", "%||%", "REPEAT", "LENGTH", "STRING_AGG", "GREATEST", "LIST_EXTRACT", "LOG10", "LOG2", "STRING_SPLIT_REGEX", "FLOOR", "FMOD", "FDIV"))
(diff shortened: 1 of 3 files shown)


R/backend-dbplyr__duckdb_connection.R
```diff
@@ -160,7 +160,7 @@ sql_translation.duckdb_connection <- function(con) {
       regexpr = function(p, x) {
         build_sql("(CASE WHEN REGEXP_MATCHES(", x, ", ", p, ") THEN (LENGTH(LIST_EXTRACT(STRING_SPLIT_REGEX(", x, ", ", p, "), 0))+1) ELSE -1 END)")
       },
-      round = function(x, digits) sql_expr(ROUND(!!x, CAST(ROUND((!!digits), 0L) %AS% INTEGER))),
+      round = function(x, digits = 0) sql_expr(ROUND_EVEN(!!x, CAST(ROUND((!!digits), 0L) %AS% INTEGER))),
```
krlmlr, line 163: Do we want a test for calling `round()` with a single argument?

krlmlr (commented):
Thanks, great!

krlmlr:
Thanks! Let's add that missing test later.

PR #158, opened 2024-05-05
chore: `n_distinct()` followup

lschneiderbauer:
adress remaining comments from pull request #122 

- check for missing dots
- add an integration test that uses the window function

### DIFF
--- a/R/backend-dbplyr__duckdb_connection.R
+++ b/R/backend-dbplyr__duckdb_connection.R
@@ -79,6 +79,12 @@ duckdb_grepl <- function(pattern, x, ignore.case = FALSE, perl = FALSE, fixed = 
 
 duckdb_n_distinct <- function(..., na.rm = FALSE) {
   sql <- pkg_method("sql", "dbplyr")
+  check_dots_unnamed <- pkg_method("check_dots_unnamed", "rlang")
+
+  if (missing(...)) {
+    cli::cli_abort("{.arg ...} is absent, but must be supplied.")
+  }
+  check_dots_unnamed()
 
   if (!identical(na.rm, FALSE)) {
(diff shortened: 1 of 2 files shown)


R/backend-dbplyr__duckdb_connection.R
```diff
@@ -79,6 +79,12 @@ duckdb_grepl <- function(pattern, x, ignore.case = FALSE, perl = FALSE, fixed =
 
 duckdb_n_distinct <- function(..., na.rm = FALSE) {
   sql <- pkg_method("sql", "dbplyr")
+  check_dots_unnamed <- pkg_method("check_dots_unnamed", "rlang")
+
+  if (missing(...)) {
+    stop("`...` is absent, but must be supplied.")
+  }
+  check_dots_unnamed()
```
krlmlr, line 87: Can this be just:

```suggestion
  if (missing(...)) {
    stop("`...` is absent, but must be supplied.")
  }
  rlang::check_dots_unnamed()
```

Similarly with `dbplyr::sql()` ?

krlmlr:
Thanks! Happy to review a follow-up using more `::` .

R/backend-dbplyr__duckdb_connection.R
lschneiderbauer: Since rlang is declared neither in Imports nor in Depends in the DESCRIPTION file I believe R CMD CHECK will complain if we use it like this.
That is actually the reason I used `pkg_method()`. I also thought that's exactly the reason `pkg_method()` exists in the first place. Is this incorrect?

krlmlr: `R CMD check` doesn't complain about `::` for suggested packages. The `pkg_method()` trick might be necessary for non-exported methods.

PR #160, opened 2024-05-07
handle data frames with 0 columns in `rel_from_df()`

romainfrancois:
related to https://github.com/hannes/duckdb-rfuns/pull/91#discussion_r1592100105


hannes:
Thanks, seems like we have to revisit the question of whether we want to allow zero-column result sets, as I recall last time we discussed this we decided not to have those but circumstances may have changed.

Tmonster:
Found this document I wrote when we first encountered the question.

https://docs.google.com/document/d/1oKaCP62b8tjkOooX7rNUVGwF3waM3PhBMzuB-PLxqtc/edit#heading=h.lq8f1uo0vi62

It seems like for aggregations with no groups, results are inconsistent between DuckDB/sqlite and dplyr

krlmlr:
Thanks, Tom. Aggregation with no groups is different, this is about data frames without columns.

PR #161, opened 2024-05-08
`expr_constant(, typed_logical_null = )`

romainfrancois:
This is related to https://github.com/hannes/duckdb-rfuns/pull/91#discussion_r1593002667

This allows `expr_constant()` to construct a NULL constant of type LOGICAL which wasn't possible before. 

Perhaps a better alternative would be to allow `expr_constant(NULL)` to construct a `SQLNULL` and let `expr_constant(NA)` to construct a `Value(LogicalType::BOOLEAN)` ?

`expr_constant(NULL)` is currently an error

### DIFF
--- a/src/relational.cpp
+++ b/src/relational.cpp
@@ -60,11 +60,11 @@ external_pointer<T> make_external_prot(const string &rclass, SEXP prot, ARGS &&.
 	return make_external<ColumnRefExpression>("duckdb_expr", names);
 }
 
-[[cpp11::register]] SEXP rapi_expr_constant(sexp val) {
+[[cpp11::register]] SEXP rapi_expr_constant(sexp val, bool typed_logical_null) {
 	if (LENGTH(val) != 1) {
 		stop("expr_constant: Need value of length one");
 	}
-	return make_external<ConstantExpression>("duckdb_expr", RApiTypes::SexpToValue(val, 0, false));
+	return make_external<ConstantExpression>("duckdb_expr", RApiTypes::SexpToValue(val, 0, typed_logical_null));
 }
 
 [[cpp11::register]] SEXP rapi_expr_function(std::string name, list args, list order_bys, list filter_bys) {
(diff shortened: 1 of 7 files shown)


tests/testthat/test-relational.R
```diff
@@ -76,6 +76,13 @@ test_that("we can create various expressions and don't crash", {
   expect_true(TRUE)
 })
 
+test_that("we can create a constant NA with LOGICAL type", {
+  rel1 <- rel_from_df(con, data.frame(x = 42))
+  exprs <- list(
+    y = expr_constant(NA, TRUE)
+  )
+  expect_equal(rel_to_altrep(rel_project(rel1, exprs))[, ], NA)
```
romainfrancois, line 84: Is there a more straightforward way to convert a constant to a SEXP rather that having to round trip through a data frame ?

src/relational.cpp
```diff
@@ -60,11 +60,11 @@ external_pointer<T> make_external_prot(const string &rclass, SEXP prot, ARGS &&.
 	return make_external<ColumnRefExpression>("duckdb_expr", names);
 }
 
-[[cpp11::register]] SEXP rapi_expr_constant(sexp val) {
+[[cpp11::register]] SEXP rapi_expr_constant(sexp val, bool typed_logical_null) {
```
romainfrancois, line 63: I believe this should instead not have the `typed_logical_null` argument, handle `val == R_NilValue` upfront, and use `RApiTypes::SexpToValue(val, 0, true)`

tests/testthat/test-relational.R
krlmlr: For our use case, this is best. You could add a wrapper function in this package?

src/relational.cpp
krlmlr: How so? I just changed this to `typed_logical_null := false` to support a special case in duckdb. I need to better understand the original motivation.

krlmlr (commented):
Thanks, Romain!

tests/testthat/test-relational.R
romainfrancois: A function that turns a constant into an R value ?

krlmlr:
Obsolete.

PR #164, opened 2024-05-08
feat: With `duckdb(environment_scan = TRUE)`, data frame objects are available as views in duckdb SQL queries

### LINKED ISSUE #140: Execute queries without registering Data Frames / Arrow Tables
From duckdb/duckdb#6771

It is convenient in the Python client to specify the target of a query without having to register `pandas.DataFrame`, etc., so it would be nice to have the same functionality in R.

hannes:
```R
duckdb:::sql('from iris')
```

now "just" works

CC @krlmlr 

Closes #140.


hannes:
Needs still tests and some protect stuff I guess

src/register.cpp
```diff
@@ -46,6 +46,41 @@ using namespace duckdb;
...
+		if (df != R_UnboundValue) {
+			break;
+		}
+		rho = ENCLOS(rho);
+	}
+	if (!df) {
+		return nullptr;
+	}
+	if (TYPEOF(df) == PROMSXP) {
+		df = Rf_eval(df, rho);
```
krlmlr, line 69: cpp11 has a global protection mechanism, but how to control the lifetime?

src/register.cpp
```diff
@@ -46,6 +46,41 @@ using namespace duckdb;
...
+	}
+	if (!Rf_inherits(df, "data.frame")) {
+		return nullptr;
+	}
+	// TODO: do utf conversion
+	auto table_function = make_uniq<TableFunctionRef>();
+	vector<duckdb::unique_ptr<ParsedExpression>> children;
+	children.push_back(
+	make_uniq<ConstantExpression>(Value::POINTER((uintptr_t)df)));
```
krlmlr, line 78: Do we have expressions that can have an arbitrary smartptr as payload?

src/register.cpp
```diff
@@ -46,6 +46,41 @@ using namespace duckdb;
...
+		return nullptr;
+	}
+	if (TYPEOF(df) == PROMSXP) {
+		df = Rf_eval(df, rho);
+	}
+	if (!Rf_inherits(df, "data.frame")) {
+		return nullptr;
+	}
+	// TODO: do utf conversion
+	auto table_function = make_uniq<TableFunctionRef>();
```
krlmlr, line 75: Or perhaps subclass `TableFunctionRef` ?

krlmlr (commented):
Thanks, nice!

src/register.cpp
hannes: indeed lifetime is a problematic issue. For example, once someone creates a view of this and then deletes the df and gc runs we might be in trouble.

src/register.cpp
hannes: no

src/register.cpp
krlmlr: IIUC, support for lifetime control has landed via duckdb/duckdb#11761, which is why we now have conflicts here?

Tishj: We already had support for lifetime control, but it was broken in a couple ways

We use the `ExternalDependency` class for lifetime control, but this wasn't being kept alive long enough, which this PR aimed to address.

We changed the prototype of the replacement scan, we changed the `const string &table_name` into `ReplacementScanInput`

I have a follow up PR scheduled that removes the newly added `table_ref &TableRef` from this struct though
Instead we will collect replacement scans inside QueryRelation::Bind (virtual overload) and push these replacement scans as CTEs onto the parsed QueryNode stored inside the QueryRelation.

krlmlr: Thanks, Thijs. May I get back to you for help with lifetime control after that follow-up PR has made it into the R package? I don't find that PR in https://github.com/duckdb/duckdb/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+author%3ATishj.

krlmlr: @Tishj: how would I go about controlling the lifetime of the bound object in this PR?

krlmlr:
Sure! I'd like to make it "opt in" though, otherwise it may (and probably will) break user code.

eitsupi:
> it may (and probably will) break user code.

Is it really so?
Wouldn't such an external table be queried if there is no table of that name in the existing connection (i.e., the query would fail without this feature)?
https://github.com/duckdb/duckdb-r/issues/140#issuecomment-2075001443

NateNohling:
Is there a reprex for how should this work? I can't seem to get this to work even on duckdb_1.1.3-1

krlmlr:
There are tests.

hannes:
cc @viktorleis

PR #167, opened 2024-05-08
expr_constant(NA)

romainfrancois:
Simpler (I believe) Alternative to #161 so that we can get a `LOGICAL NULL` constant with `expr_constant(NA)` and a `SQL NULL` from `expr_constant(NULL)`


krlmlr:
Obsolete.

PR #168, opened 2024-05-12
feat: New `sort` argument to `rel_order()`

toppyy:
Related to duckdblabs/duckplyr#92

Changes:
1. `rapi_rel_order` now accepts a vector of boolean values that describe whether each expression should be sorted in ascending order
2. `rel_order` checks that the number of expressions and boolean values match or defaults to TRUE if a null-value is passed
3. tests for the relation api test the `rel_order()`, not `rapi_rel_order()`. The tests break without this change as the default value is handled in rel_order.

I will mark this as a draft due to these notes:
 - This does not work for nested expressions. For example, assume someone writes `arrange(someotherfunction(desc(column)))` in duckplyr. I don't know if this is an issue as the interpretation of such an expression in terms of sort order is a bit ambiguous(?).
    - note: this problem exists also without changes proposed here?
 - I was unsure what function should to check that `length(orders) == length(ascending)`. It is now `rel_order()`, `rapi_rel_order()`. As a result, using rapi_rel_order directly might result indexing out of array bounds. Perhaps the check should be in rapi_rel_order as the vector is indexed there?
 - Related to the previous note. Should we aim to test `rel_order()` or `rapi_rel_order`? There are examples of both for other relation expressions


src/relational.cpp
```diff
@@ -214,11 +214,16 @@ external_pointer<T> make_external_prot(const string &rclass, SEXP prot, ARGS &&.
...
 }
 
-[[cpp11::register]] SEXP rapi_rel_order(duckdb::rel_extptr_t rel, list orders) {
+[[cpp11::register]] SEXP rapi_rel_order(duckdb::rel_extptr_t rel, list orders, r_vector<r_bool> ascending) {
 	vector<OrderByNode> res_orders;
 
+	OrderType order_type;
+	size_t i = 0;
+
 	for (expr_extptr_t expr : orders) {
```
krlmlr, line 223: @Antonov548: What's the idiomatic C++ way to iterate over two "parallel" vectors of the same length, `orders` and `ascending` ?

krlmlr (commented):
Thanks, looks great! Triggering checks.

krlmlr:
Testing `rel_order()` is fine, `rapi_rel_order()` is low-level and will never be exported. The `arrange(f(desc(x)))` issue is not much of a problem -- duckplyr will only translate `desc()` if it's the outermost expression and fall back to dplyr otherwise.

src/relational.cpp
Antonov548: As I know there is no specific function in STL to do this. Usually it's range based for loop with `:` for single iterable objects and index based loop for multiple vectors.

krlmlr: Can we use `zip_view()` or similar?

https://codereview.stackexchange.com/q/283638/10691

krlmlr:
Thanks!

PR #175, opened 2024-05-30
fix: Correct usage of `win_current_group()` instead of `win_current_order()` in SQL translation

### LINKED ISSUE #173: `n_distinct()` does not work with non-trivial windows
I cannot run a reprex just now, but the following does not take into account the partition column `x` which is incorrect.

```R
library(dplyr)
library(duckdb)
library(DBI)

df <- tibble(x = c(1, 1, 2), y = c(1, 1, 2))

con <- DBI::dbConnect(duckdb::duckdb())
duckdb::duckdb_register(con, "df", df)

tbl(con, "df") |> mutate(z = n_distinct(y), .by = x) |> show_query()
```

I am planning to provide a fix on the upcoming weekend.

lschneiderbauer:
Fixes typo and adds tests

- fixes #173 
- also accidently fixes `str_flatten()` in a windowed context


krlmlr:
Thanks!

PR #185, opened 2024-06-18
Add install.libs script to remove debug symbols

hannes:
Custom install script, somehow makes CRAN not ship debug symbols ^^
thanks @gaborcsardi


krlmlr:
Thanks. This seems to be a copy of the code in WRE in https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Package-subdirectories . Do we need to adapt this to achieve the desired behavior? Why does adding this file reduce the size of the shared library?

krlmlr:
Thanks, now part of duckdb 1.0.0-2, on CRAN.

PR #457, opened 2024-10-06
feat: Add comparison expression to relational api

toppyy:
Related to tidyverse/duckplyr#172

Add comparison expressions for duckdplyr

PR #462, opened 2024-10-15
test: Add tests for comparison expression

toppyy:
Adds tests for comparison expression in the relational api (added in #457)


krlmlr:
Thanks! I also added a more complete test to the end of that file. Do we need to move code to bring these tests closer together?

PR #530, opened 2024-10-27
feat: xz-compress duckdb sources in the tarball

krlmlr:


.Rbuildignore
```diff
@@ -43,3 +43,5 @@ _files$
 ^compile_commands\.json$
 ^\.cache$
 ^tests/testthat/helper-rethrow\.R$
+^cleanup$
+^cleanup\.win$
```
eitsupi, line 47: Is it correct that these are here?

eitsupi (commented):
I do not know if the cleanup script is the right place to do this.

.Rbuildignore
krlmlr: I think so, the tarball has the `.xz` file and doesn't need this procedure.

krlmlr: `cleanup` is run as part of `R CMD build .` in the right location at the right time. Seems legit, but 🤷 .

hannes:
fine by me

hannes (approved):

Issue #612, opened 2024-11-26
Compressed source seems to break CI workflows that install duckdb


andrewGhazi:
Hi there -- I have a Carpentries training module that depends on `duckdb` and installs the package on a Github runner image to generate the online version. However the relatively recent addition of the xz compressed source directory seems to be breaking the Github actions that orchestrate this process. See the log here: 

https://github.com/carpentries-incubator/bioc-scrnaseq/actions/runs/12036979601/job/33559404345#step:6:2590

By the time it gets to `tar cvJf duckdb.tar.xz duckdb` that directory seems to be gone. I guess it's deleted early somehow, maybe it's [the cleanup script](https://github.com/duckdb/duckdb-r/blob/main/cleanup) somehow running *before* configure or something? It's hard to tell exactly what's going on, but it does seem to be a relatively recent bug as another workflow from a few weeks ago that used `duckdb_1.1.2` from CRAN succeeded.

Maybe this is my workflow's problem, but I thought I would flag the issue here since it was pretty tricky to track down. It would be nice not to have to pin the module to an old `duckdb` version.

krlmlr:
Thanks. It looks like we need to make `./cleanup` reentrant: check if `.tar.xz` exists, if yes, do nothing.

andrewGhazi:
Hi there, unfortunately the update in 1.1.13-1 doesn't seem to have fixed the issue for me. I wish I could be of more help but I can't tell why the compression command is getting run given the update in #634 ... The error log is pretty much the same as last time:

```
  
  * installing *source* package ‘duckdb’ ...
  ** package ‘duckdb’ successfully unpacked and MD5 sums checked
  ** using staged installation
  + [ -f duckdb.tar.xz ]
  + which xz
  + [ -d .git ]
  + cd src
  + tar cvJf duckdb.tar.xz duckdb
  tar: duckdb: Cannot stat: No such file or directory
  tar: Exiting with failure status due to previous errors
  + dirname ./configure
  + cd ./src
  + [ -f  ]
  + cp include/to-tar.mk Makevars.duckdb
  + [ -f duckdb.tar.xz ]
  + tar xJf duckdb.tar.xz
  ** libs
  using C++ compiler: ‘g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0’
  using C++17
  make: *** No rule to make target 'duckdb/ub_src_catalog.o', needed by 'duckdb.tar'.  Stop.
[...]

PR #634, opened 2024-12-07
fix: Make `./cleanup` script reentrant

Antonov548:
Skip cleanup if archive is already exists. 

Closes https://github.com/duckdb/duckdb-r/issues/612.


krlmlr:
Thanks!

PR #707, opened 2024-12-24
chore: Set value returned by `PRAGMA user_agent` to r-dbi

elefeint:
R clients using `PRAGMA user_agent` historically fell back on the default "cpp" value of `duckdb_api`. Example output: 
```
duckdb/v1.1.4-dev2005(linux_amd64) cpp
```

This PR sets an R specific `r-dbi` value. I used `r-dbi` as an analogy to `jdbc`, which indicates both the language and the API. Would another `duckdb_api` string be more appropriate?

Example output with this PR:
```
duckdb/v1.1.4-dev2005(linux_amd64) r-dbi
```


elefeint:
@krlmlr I merged from main; could you trigger the test run when you get a chance?

krlmlr:
Thanks!

PR #1063, opened 2025-02-28
fix: Fix support for the connections pane in RStudio and Positron

dfalbel:
This PR fixes support for RStudio/Positron connections pane.
Should be able to navigate through databases, including those that have a few different schemas.

CleanShot 2025-02-28 at 12 27 05

See also: https://github.com/posit-dev/positron/issues/6517


krlmlr:
Thanks!

PR #1090, opened 2025-04-07
feat: Add support for duckdb arrays in R

### LINKED ISSUE #102: New array data type in version 0.10.0 can't be returned to R
```r
con <- dbConnect(duckdb())
dbGetQuery(con, "select array_value(1,2,3)")
#> Error: rapi_prepare: Unknown column type for prepare: INTEGER[3]
```

joakimlinde:
Limited to one-dimensional duckdb arrays for now. 

Closes #102

### DIFF
--- a/src/reltoaltrep.cpp
+++ b/src/reltoaltrep.cpp
@@ -239,7 +239,7 @@ struct AltrepVectorWrapper {
 			idx_t dest_offset = 0;
 			for (auto &chunk : res->Collection().Chunks()) {
 				SEXP dest = transformed_vector.data();
-				duckdb_r_transform(chunk.data[column_index], dest, dest_offset, 1, chunk.size(), false);
+				duckdb_r_transform(chunk.data[column_index], dest, dest_offset, chunk.size(), false);
 				dest_offset += chunk.size();
 			}
 		}
--- a/src/reltoaltrep.cpp
+++ b/src/reltoaltrep.cpp
@@ -239,7 +239,7 @@ struct AltrepVectorWrapper {
 			idx_t dest_offset = 0;
 			for (auto &chunk : res->Collection().Chunks()) {
 				SEXP dest = transformed_vector.data();
-				duckdb_r_transform(chunk.data[column_index], dest, dest_offset, chunk.size(), false);
+				duckdb_r_transform(chunk.data[column_index], dest, dest_offset, 1, chunk.size(), false);
 				dest_offset += chunk.size();
 			}
 		}
--- a/src/transform.cpp
+++ b/src/transform.cpp
@@ -163,7 +163,7 @@ void duckdb_r_decorate(const LogicalType &type, const SEXP dest, bool integer64)
 		auto array_size = ArrayType::GetSize(type);
 		auto &child_type = ArrayType::GetChildType(type);
 		duckdb_r_decorate(child_type, dest, integer64);
-		// The class of a matrix and an array is implicit from 
+		// The class of a matrix and an array is implicit from
 		// the dim attribute so we don't set the class attribute.
 		// See: https://svn.r-project.org/R/trunk/src/main/attrib.c:656
 		// SET_CLASS(dest, RStrings::get().matrix_array_str);
@@ -345,7 +345,7 @@ void duckdb_r_transform(Vector &src_vec, const SEXP dest, idx_t dest_offset, idx
 		break;
 	case LogicalTypeId::BIGINT:
 		if (integer64) {
-			VectorToR<int64_t, int64_t>(src_vec, n, NUMERIC_POINTER(dest), dest_offset, dest_step_size, 
+			VectorToR<int64_t, int64_t>(src_vec, n, NUMERIC_POINTER(dest), dest_offset, dest_step_size,
 			                            NumericLimits<int64_t>::Minimum());
 			Rf_setAttrib(dest, R_ClassSymbol, RStrings::get().integer64_str);
 		} else {
--- a/src/transform.cpp
+++ b/src/transform.cpp
@@ -235,6 +235,21 @@ SEXP ToRString(const string_t &input) {
 	return Rf_mkCharLenCE(data, len, CE_UTF8);
 }
 
+static void TransformArrayVector(Vector &src_vec, const SEXP dest, idx_t dest_offset, idx_t n, bool integer64) {
+	auto array_size = ArrayType::GetSize(src_vec.GetType());
+	auto &child_type = ArrayType::GetChildType(src_vec.GetType());
+	Vector child_vector(child_type, nullptr);
+	auto matrix_nrow = (Rf_xlength(dest) / array_size);
+
+	// actual loop over rows
+	for (size_t row_idx = 0; row_idx < n; row_idx++) {
+		size_t offset = (row_idx * array_size);
+		size_t end = offset + array_size;
+		child_vector.Slice(ArrayVector::GetEntry(src_vec), offset, end);
+		duckdb_r_transform(child_vector, dest, dest_offset + row_idx, matrix_nrow, array_size, integer64);
+	}
+}
+
 void duckdb_r_transform(Vector &src_vec, const SEXP dest, idx_t dest_offset, idx_t dest_step_size, idx_t n, bool integer64) {
 	if (src_vec.GetType().GetAlias() == R_STRING_TYPE_NAME) {
 		ptrdiff_t sexp_header_size = (data_ptr_t)DATAPTR_RO(R_BlankString) - (data_ptr_t)R_BlankString;
@@ -445,19 +460,7 @@ void duckdb_r_transform(Vector &src_vec, const SEXP dest, idx_t dest_offset, idx
 		break;
 	}
 	case LogicalTypeId::ARRAY: {
-		// figure out the total and max element length of the list vector child
-		auto array_size = ArrayType::GetSize(src_vec.GetType());
-		auto &child_type = ArrayType::GetChildType(src_vec.GetType());
-		Vector child_vector(child_type, nullptr);
-		auto matrix_nrow = (Rf_xlength(dest) / array_size);
-
-		// actual loop over rows
-		for (size_t row_idx = 0; row_idx < n; row_idx++) {
-			size_t offset = (row_idx * array_size); 
-			size_t end = offset + array_size;
-			child_vector.Slice(ArrayVector::GetEntry(src_vec), offset, end);
-			duckdb_r_transform(child_vector, dest, dest_offset + row_idx, matrix_nrow, array_size, integer64);
-		}
+		TransformArrayVector(src_vec, dest, dest_offset, n, integer64);
 		break;
 	}
 	case LogicalTypeId::STRUCT: {
--- a/src/transform.cpp
+++ b/src/transform.cpp
@@ -7,13 +7,13 @@ using namespace duckdb;
 
 // converter for primitive types
 template <class SRC, class DEST>
-static void VectorToR(Vector &src_vec, size_t count, void *dest, uint64_t dest_offset, uint64_t dest_step_size, DEST na_val) {
+static void VectorToR(Vector &src_vec, size_t count, void *dest, uint64_t dest_offset, DEST na_val) {
 	auto src_ptr = FlatVector::GetData<SRC>(src_vec);
 	auto &mask = FlatVector::Validity(src_vec);
 	auto dest_ptr = ((DEST *)dest) + dest_offset;
 	for (size_t row_idx = 0; row_idx < count; row_idx++) {
 		dest_ptr[0] = !mask.RowIsValid(row_idx) ? na_val : src_ptr[row_idx];
-		dest_ptr += dest_step_size;
+		dest_ptr += 1;
 	}
 }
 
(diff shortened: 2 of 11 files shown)


joakimlinde:
Usage example:

``` r
library(tidyverse)
library(duckdb)

con <- dbConnect(duckdb())
on.exit(dbDisconnect(con, shutdown = TRUE))

dbExecute(con, "CREATE TABLE tbl (a INTEGER, b INTEGER[3])")
dbExecute(con, "INSERT INTO tbl VALUES (10, [1, 4, 7])")
dbExecute(con, "INSERT INTO tbl VALUES (11, [2, 5, 8])")
dbExecute(con, "INSERT INTO tbl VALUES (12, [3, 6, 9])")

df <- as_tibble(dbGetQuery(con, "FROM tbl"))
df
#> # A tibble: 3 × 2
#>       a b[,1]  [,2]  [,3]
#>   <int> <int> <int> <int>
#> 1    10     1     4     7
#> 2    11     2     5     8
#> 3    12     3     6     9

df %>% filter(a == 11)
#> # A tibble: 1 × 2
#>       a b[,1]  [,2]  [,3]
#>   <int> <int> <int> <int>
#> 1    11     2     5     8

df %>% arrange(desc(a))
#> # A tibble: 3 × 2
#>       a b[,1]  [,2]  [,3]
#>   <int> <int> <int> <int>
#> 1    12     3     6     9
#> 2    11     2     5     8
#> 3    10     1     4     7
```

src/transform.cpp
```diff
@@ -7,12 +7,20 @@ using namespace duckdb;
...
+	if (dest_step_size == 1) {
+		// This one is for speed. Maybe a tiny bit faster if the optimzer can figure it out. 	
+		for (size_t row_idx = 0; row_idx < count; row_idx++) {
+			dest_ptr[row_idx] = !mask.RowIsValid(row_idx) ? na_val : src_ptr[row_idx];
+		}
+	} else {
+		for (size_t row_idx = 0; row_idx < count; row_idx++) {
+			dest_ptr[0] = !mask.RowIsValid(row_idx) ? na_val : src_ptr[row_idx];
+			dest_ptr += dest_step_size;
+		}
```
krlmlr, line 23: IIUC, the difference between the two loops is between `++` and `+= dest_step_size` ?

I'd like to look at the generated code before optimizing at this fine-grained level. My gut feeling says it won't matter because there is a lot of unpredictable branching going on due to the `?` inside the loop, and that for modern compilers/processors the `+=` is not the main bottleneck.

src/transform.cpp
```diff
@@ -50,6 +58,14 @@ SEXP duckdb_r_allocate(const LogicalType &type, idx_t nrows) {
 	case LogicalTypeId::LIST:
 	case LogicalTypeId::MAP:
 		return NEW_LIST(nrows);
+	case LogicalTypeId::ARRAY: {
+		auto array_size = ArrayType::GetSize(type);
+		auto &child_type = ArrayType::GetChildType(type);
+		if (child_type.IsNested())
+			cpp11::stop("rapi_execute: Arrays must not be nested.");
```
krlmlr, line 65: "Nested arrays cannot be returned to R as column data" or similar?

What's a good wording here? DuckDB supports it, only we don't yet.

krlmlr (commented):
Thanks, got halfway through, will follow up. Why are the checks failing?

joakimlinde:
The tests are not failing, they need to be approved to run. I see "1 workflow awaiting approval. This workflow requires approval from a maintainer." More info [here](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks).

src/transform.cpp
```diff
@@ -7,12 +7,20 @@ using namespace duckdb;
...
 	auto dest_ptr = ((DEST *)dest) + dest_offset;
-	for (size_t row_idx = 0; row_idx < count; row_idx++) {
-		dest_ptr[row_idx] = !mask.RowIsValid(row_idx) ? na_val : src_ptr[row_idx];
+	if (dest_step_size == 1) {
+		// This one is for speed. Maybe a tiny bit faster if the optimzer can figure it out. 	
+		for (size_t row_idx = 0; row_idx < count; row_idx++) {
+			dest_ptr[row_idx] = !mask.RowIsValid(row_idx) ? na_val : src_ptr[row_idx];
+		}
+	} else {
+		for (size_t row_idx = 0; row_idx < count; row_idx++) {
```
krlmlr, line 20: Shouldn't this be a nested loop? The loop over `row_idx` iterates over rows, and in each row, we need to populate a matrix row? What am I missing?

krlmlr (commented):
Thanks. Are we covering ingestion (going from R to DuckDB) in a separate PR?

joakimlinde:
We can do ingestion too? I missed that. I'll add it.

krlmlr:
Let's do it in a separate PR then.

src/reltoaltrep.cpp
```diff
@@ -426,7 +426,7 @@ size_t DoubleToSize(double d) {
 		R_SetExternalPtrTag(ptr, RStrings::get().duckdb_vector_sym);
 
 		cpp11::sexp vector_sexp = R_new_altrep(LogicalTypeToAltrepType(column_type), ptr, R_NilValue);
-		duckdb_r_decorate(column_type, vector_sexp, false);
+		duckdb_r_decorate(column_type, vector_sexp, DoubleToSize(n_rows), false);
```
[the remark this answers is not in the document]
krlmlr: It's in the signature, missed that.

tests/testthat/test-array.R
```diff
@@ -0,0 +1,197 @@
+previous_edition = edition_get()
+local_edition(3)
```
krlmlr, line 2: The `local_*()` family of functions automatically restores the old state after leaving the scope. Do we need to restore?

It's fine to repeat in each `test_that()` block, too. At some point, we migrate to edition 3 and remove all of them.

joakimlinde:
Sure, I'll add ingestion in a separate PR.

src/transform.cpp
joakimlinde: I wasn't sure how sensitive you were to performance so I provided both as an options. I'll remove the special case of 1  and let the generic one handle it.

src/transform.cpp
joakimlinde: Error messages are always hard. You want to capture everything in a short sentence. I like your proposed text and will update.

src/transform.cpp
```diff
@@ -150,6 +166,22 @@ void duckdb_r_decorate(const LogicalType &type, const SEXP dest, bool integer64)
 	case LogicalTypeId::LIST:
 	case LogicalTypeId::MAP:
 		break; // no extra decoration required, do nothing
+	case LogicalTypeId::ARRAY: {
+		auto array_size = ArrayType::GetSize(type);
```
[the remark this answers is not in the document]
joakimlinde: Probably a tab vs space thing. Let me take a look. The only true tab setting is 8.

src/transform.cpp
```diff
@@ -150,6 +166,22 @@ void duckdb_r_decorate(const LogicalType &type, const SEXP dest, bool integer64)
...
+		auto &child_type = ArrayType::GetChildType(type);
+		if (child_type.IsNested())
+			cpp11::stop("rapi_execute: Arrays must not be nested.");
+		duckdb_r_decorate(child_type, dest, nrows, integer64);
+		// The class of a matrix and an array is implicit from 
+		// the dim attribute so we don't set the class attribute.
+		// See: https://svn.r-project.org/R/trunk/src/main/attrib.c:656
+		// SET_CLASS(dest, RStrings::get().matrix_array_str);
+		cpp11::sexp dims = NEW_INTEGER(2);
+		INTEGER(dims)[0] = nrows;
```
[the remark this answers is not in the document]
joakimlinde: That's a good point. Let me try that. It would simplify things.

src/transform.cpp
joakimlinde: `VectorToR()` takes one DuckDB array in one DuckDB record and writes it to one row in the R matrix we are building. The `for` loop you are looking for, the one that loops over all the records in the DuckDB table, is statement .cpp:181.

``` cpp
idx_t dest_offset = 0;
for (auto &chunk : result->Collection().Chunks()) {
	D_ASSERT(chunk.ColumnCount() == ncols);
	D_ASSERT(chunk.ColumnCount() == (idx_t)Rf_length(data_frame));
	for (size_t col_idx = 0; col_idx < chunk.ColumnCount(); col_idx++) {
		SEXP dest = VECTOR_ELT(data_frame, col_idx);
		duckdb_r_transform(chunk.data[col_idx], dest, dest_offset, 1, chunk.size(), integer64);
	}
	dest_offset += chunk.size();
}
```

It seems like you think each DuckDB record generates a matrix but that's not the case. It's the whole column with an array in each record that forms the matrix.

tests/testthat/test-array.R
joakimlinde: I tried that but it didn't restore it for some reason and other tests starting to complain that functions has been deprecated in the 3rd edition. That's why I had to come up with this restore-state myself workaround. But I'll take another look at it with a fresh(er) mind.

joakimlinde: I looked at it and if you put the `local_edition()` at the top of the file it does restore it properly. However, if you put it inside a test case it does not restore it automatically and you have to restore it yourself. This is what got me confused because I initially had the `local_edition()` inside the test cases and then moved it to the top.

joakimlinde:
I've updated the submission to address the comments as discussed.

src/transform.cpp
```diff
@@ -415,26 +436,41 @@ void duckdb_r_transform(Vector &src_vec, const SEXP dest, idx_t dest_offset, idx
...
+		auto &child_type = ArrayType::GetChildType(src_vec.GetType());
+		Vector child_vector(child_type, nullptr);
+		auto matrix_nrow = (Rf_xlength(dest) / array_size);
+
+		// actual loop over rows
+		for (size_t row_idx = 0; row_idx < n; row_idx++) {
+			size_t offset = (row_idx * array_size); 
+			size_t end = offset + array_size;
+			child_vector.Slice(ArrayVector::GetEntry(src_vec), offset, end);
+			duckdb_r_transform(child_vector, dest, dest_offset + row_idx, matrix_nrow, array_size, integer64);
```
krlmlr, line 459: I now see that `duckdb_r_transform()` is employed in two ways:

- Iterating over rows
- Iterating over all values in an array row

Essentially, we're implementing matrix transposition here, which I remember to be notoriously cache-inefficient.

Because of this, we can as well use a temporary buffer here (that saves us the `dest_step_size` argument everywhere) and distribute the values from that temporary buffer to the correct indices here. As a next step, we can then switch to process in 16x16 blocks or so; the buffer will then fit in the L1 cache, and the extra copy won't matter that much.

Would you like to go the extra mile here?

krlmlr (commented):
Thanks. IIUC, getting rid of the `idx_t dest_step_size` argument to `duckdb_r_transform()` requires a temporary buffer. I'm fine with that because I think the true optimization potential is in improving cache efficiency.

src/transform.cpp
```diff
@@ -150,6 +159,20 @@ void duckdb_r_decorate(const LogicalType &type, const SEXP dest, bool integer64)
...
+		// The class of a matrix and an array is implicit from 
+		// the dim attribute so we don't set the class attribute.
+		// See: https://svn.r-project.org/R/trunk/src/main/attrib.c:656
+		// SET_CLASS(dest, RStrings::get().matrix_array_str);
+		cpp11::sexp dims = NEW_INTEGER(2);
+		INTEGER(dims)[0] = (Rf_xlength(dest) / array_size);
+		INTEGER(dims)[1] = array_size;
+		Rf_setAttrib(dest, RStrings::get().dim_sym, dims);
+		}
+		break;
```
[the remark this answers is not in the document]
joakimlinde: I see what you mean. Let me fix that.

src/transform.cpp
joakimlinde: I would like to get basic functionality in place first and then do performance optimizations based on measurements. Then I can rewrite the code and make sure I've improved performance instead of making it worse.

krlmlr: My main concern with this PR is that we add an argument to a function that is called 10+ times, and we need to thread through that argument, just to cater for the new edge case. It seems that, if we go with a buffer, we can simplify a lot and the diff of the PR gets much smaller. This seems consistent with getting in basic functionality first -- I need to balance this with code simplicity. I'm happy to put in that change, also happy to review yours.

joakimlinde:
I've fixed the consistency issue and re-submitted.

src/transform.cpp
joakimlinde: Joking aside, I understand and appreciate your concern regarding performance. It's a valid concern and it stems from how matrices are laid out in memory in R. A matrix row is scattered all over memory. Each element is nrows apart. Where we differ is that I don't think you can work around this problem — it's intrinsic to R matrices. Either your reads are going to be scattered or your writes are going to be scattered — pick your poison. This is a valid argument for using lists instead of matrices in R to represent DuckDB arrays.

krlmlr:
Thanks!

PR #1125, opened 2025-04-25
feat: Add `dbConnect(array = "none")`, pass `array = "matrix"` to enable conversion of array columns

joakimlinde:


joakimlinde:
Seems like the 'arrow' package is not available on R 4.1: `Package suggested but not available for checking: ‘arrow’`

R/dbConnect__duckdb_driver.R
```diff
@@ -55,7 +58,8 @@ dbConnect__duckdb_driver <- function(
   timezone_out = "UTC",
   tz_out_convert = c("with", "force"),
   config = list(),
-  bigint = "numeric"
+  bigint = "numeric",
+  array = "none"
```
krlmlr, line 62: I don't think `bigint` belongs here, this should be deprecated at some point. Let's not add `"array"` here. Should have been clearer about that in the code, add a comment to that effect now?

krlmlr (commented):
Thanks. The R 4.1 check failures are on me.

Do you think it's useful to split this into two PRs, and start with a refactoring PR that replaces `integer64` with `convert_opts` everywhere? This would be easier to reason about later. I'm not sure if it's really worth investing that time now -- we can always do it after the fact if needed.

joakimlinde:
The PR is fine as it is. Don't worry about it.

R/dbConnect__duckdb_driver.R
krlmlr: I'm mistaken, this is where `array = "none"` belongs. For some reason I thought this is the `duckdb()` function.

krlmlr:
Thanks!

PR #1150, opened 2025-05-13
feat: Add ingestion of matrices

joakimlinde:


joakimlinde:
Looking at R API issue...

src/scan.cpp
```diff
@@ -104,9 +107,84 @@ void AppendListColumnSegment(const RType &rtype, SEXP *source_data, idx_t sexp_o
...
+		                                                   nrows, ncols, sexp_offset, child_vector, count);
+		break;
+
+	case RType::INTEGER: //INTSXP
+		AppendMatrixSegmentAtomic<int, int, RIntegerType>(INTEGER_POINTER(source_data),
+		                                                  nrows, ncols, sexp_offset, child_vector, count);
+		break;
+
+	case RType::INTEGER64: //REALSXP
+		AppendMatrixSegmentAtomic<int64_t, int64_t, RInteger64Type>(((int64_t *)DATAPTR(source_data)),
```
krlmlr, line 150: Should this be `DATAPTR_RO()` here and elsewhere?

src/utils.cpp
```diff
@@ -111,6 +111,9 @@ static void AppendColumnSegment(SRC *source_data, Vector &result, idx_t count) {
 }
 
 R_len_t RApiTypes::GetVecSize(RType rtype, SEXP coldata) {
+	if (rtype.id() == RTypeId::MATRIX) {
```
krlmlr, line 114: Should this check and the `while` loop be swapped?

krlmlr (commented):
Thanks. Is there a good way to make this feature opt-in, based on `ConvertOpts` ?

I see this adds some support for complex numbers. Could this have been a separate PR? Same question regarding opt-in behavior.

src/scan.cpp
joakimlinde: Yes. Fixed.

joakimlinde:
> Is there a good way to make this feature opt-in, based on `ConvertOpts` ?

No, I don't think so. Here is why. If the user gives me a matrix I only have two options. Either reject it as "I don't know what to do with an array" or slice up the matrix into arrays and push that into the database. The intent is clear from the user - he/she is giving me a matrix and there is only one constructive thing I can do with it. If the user had given me lists instead it would have been different. In that case I would have to know what to do with them, either create lists or arrays in the database. In that case the user would have to give me additional direction through options. 

> I see this adds some support for complex numbers. Could this have been a separate PR? Same question regarding opt-in behavior.

I added support so I can generate an error message saying that complex numbers are not supported. I could do it in a different way by looking at the SEXP type instead if that works better for you. It would break the pattern in scan.cpp though — look a little bit as a quick-and-dirty.

src/utils.cpp
joakimlinde: Sure, I can swap them. You mean that if I have a matrix inside a structure it would work better? Not sure if I've thought about that case... Is that a case we want to support?

joakimlinde:
The Ubuntu-22.04 (4.2) check fails. Seems like the snapshot files didn't make it across to Ubuntu. Can I ask you to manually rerun the test?

krlmlr:
Thanks!

Issue #630, opened 2024-12-06
Save query without loading the data into R environment


hope-data-science:
I have a pretty large duckdb file and I use a query to generate a larger-than-memory data. How can I save this results elegantly without using `compute` to load the data into R environment (which might not be possible)?

krlmlr:
Thanks, we need to document this better, and/or provide utilities.

For now, you could use `COPY TO` like this:

```r
data_sql <- "..."
output_sql <- paste0("COPY (", data_sql, ") TO 'out.parquet' (FORMAT PARQUET, ...)")
dbExecute(con, output_sql)
```

Upstream docs: https://duckdb.org/docs/sql/statements/copy#copy--to

hope-data-science:
FYI, currently I am using:
```
## to parquet
duckdb_sink_parquet = \(conn,sql,path) {
  DBI::dbExecute(conn = conn,
                 statement =  stringr::str_glue("COPY ({sql})
          TO '{path}' (FORMAT PARQUET)"))
}

## to a table in the current database (duckdb)
duckdb_sink_table = \(conn,sql,table_name){
  DBI::dbExecute(conn = conn,
                 statement =  stringr::str_glue("CREATE TABLE {table_name} AS {sql}"))
}
```

krlmlr:
Thanks. This is out of scope for DBI, could be in scope for duckdb of for a utility package on top of duckdb.

hope-data-science:
Another practice I tried recently is:
```
tbl(duckdb_connecntion, "table_name") %>%
    sql_render() %>%
    db_save_query(con = duckdb_connection,sql = ., name = "new_table_name",temporary = FALSE)
```
I think we really need this sort of functions to implement out-of-memory computation.

Thanks.

lgaborini:
If you have a lazy table `dt` that can't be materialized in-memory, I have also just discovered this:

```r
dt |>
  compute(
    name = "new_table_name",
    overwrite = TRUE,
    temporary = FALSE
  )
```

krlmlr:
This works much better in the upcoming duckplyr 1.1.0 . The hand-rolled variants shown above remain valid.

krlmlr:
Works in duckplyr:

``` r
library(tidyverse)
library(duckdb)
#> Loading required package: DBI

con <- dbConnect(duckdb::duckdb())

dbWriteTable(con, "my_tbl", data.frame(a = 1))

dbplyr_tbl <- tbl(con, "my_tbl")

[...]

PR #1177, opened 2025-05-25
feat: add two utility functions export_parquet and create_view and a vignette on how to materialize data.

### LINKED ISSUE #207: What is the canonical way to write parquet to disk using duckdb and dbplyr without collecting first?
Scenario is working with very large (many GB) csv files and wanting to save them to parquet files after processing without reading/collecting them entirely into memory. 

I know how to do this with arrow: 

```r
library(arrow)

open_dataset("path/to/csv", format = "csv") |> 
  do_something() |>
  write_dataset("path/to/export.parquet", format = "parquet") 
```

I also know how to do this with duckdb and SQL:

```sql
COPY
    (SELECT * FROM 'path/to/csv.csv')
    TO  'path/to/export.parquet'
    (FORMAT 'parquet')
```

But I would like to figure out how to do this with duckdb and dbplyr. Context is that members of my team are unfamiliar with SQL, and are comfortable with dplyr syntax.

The closest I've got so far is:

```r
library(arrow)
library(duckdb)

con <- dbConnect(duckdb())

tbl(con, "read_csv('path/to/csv.csv')") |>  # see note below
  do_something() |>
  to_arrow() |>
  write_dataset("path/to/export", format = "parquet") 
```
`note`: in my testing you have to pass `tbl()` the read function as well as the file path as a text string, even if the file path contains the csv or parquet extension in it. Is this where `tbl_file()` comes in? if so, how do you pass `tbl_file`  parameters to the read_csv function, such as which columns are dates? (as mentioned in #159, it's unclear how to use duckdb-specific dbplyr functions from the duckdb documentation). 

I'm a bit nervous about my workaround of using `to_arrow()` because one of the reasons we're starting as a team to use duckdb in preference to arrow is that arrow's auto-detection of schemas from csvs is not anywhere near as good as duckdb and it's very fussy and slow at parsing dates. I've also noticed how arrow interprets empty strings differs from duckdb (arrow leaves them as "", duckdb makes them NA). And I guess I'm cautious about introducing a step that might complicate matters and make it unclear what type casting has occurred by transferring between the two libraries. 

I've also tried various versions of `copy_to` and `db_copy_to` with "temporary = F". I've managed to create an in-memory table called " 'path/to/export.parquet'     (FORMAT 'parquet')" (!) but not actually save anything to disk. 

If I should ask this question elsewhere, please let me know.

nbc:
This PR add two utility functions :

* export_parquet uses COPY TO to export a parquet file from a tbl_lazy
* create_view creates a view based on a tbl_lazy

The vignette explains how to materialize data using these two functions, as well as with the lesser-known dbplyr::compute() function.

Fixes #207, #630


krlmlr:
Thanks for the effort. The code looks good, I'm still not ready to assume maintenance for it.

I labeled the issues as "help wanted" before duckplyr 1.1.0. Writing to Parquet works there:

``` r
library(tidyverse)
library(duckdb)
#> Loading required package: DBI

con <- dbConnect(duckdb::duckdb())

dbWriteTable(con, "my_tbl", data.frame(a = 1))

dbplyr_tbl <- tbl(con, "my_tbl")

dbplyr_tbl %>%
  duckplyr::as_duckdb_tibble() |>
  duckplyr::compute_parquet("my_tbl.parquet")
#> # A duckplyr data frame: 1 variable
#>       a
#>   <dbl>
#> 1     1

duckplyr_parquet <-
  duckplyr::read_parquet_duckdb("my_tbl.parquet")
duckplyr_parquet
#> # A duckplyr data frame: 1 variable
#>       a
#>   <dbl>
#> 1     1
```

<sup>Created on 2025-06-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

As for creating a view, dbplyr or a related package is a better fit, this functionality seems general enough to work across databases.

I'd like to keep the package vignette-free for shorter build times. Otherwise, each `R CMD build .` will have to install the package.

Happy to see this code thrive elsewhere!

PR #1588, opened 2025-09-28
add `n_distinct(..., na.rm = TRUE)` support for multiple passed columns.

lschneiderbauer:
My previous contribution only supported `n_distinct(..., na.rm = TRUE)` with for a single column, because duckdb's `FILTER` clause was not supported in a windows context. This changed with duckdb v1.2.

This PR adds `na.rm = TRUE` support for multiple columns by using duckdb's `struct` and using `FILTER` to filter for non-null values before counting distinct elements.

I also adjusted the tests to simply compare against the dplyr result of `n_distinct()` instead of testing against explicit values to make sure the semantics is identical.

### DIFF
--- a/R/backend-dbplyr__duckdb_connection.R
+++ b/R/backend-dbplyr__duckdb_connection.R
@@ -77,21 +77,14 @@ duckdb_n_distinct <- function(..., na.rm = FALSE) {
   }
   check_dots_unnamed()
 
+  # https://duckdb.org/docs/sql/data_types/struct.html#creating-structs-with-the-row-function
+  str_struct <- paste0("row(", paste0(list(...), collapse = ", "), ")")
+
   if (!identical(na.rm, FALSE)) {
-    cols <- list(...)
+    str_null_check <- paste0(paste0(list(...), " IS NOT NULL"), collapse = " AND ")
 
-    # check for more than one vector argument: only one vector is supported
-    # Why not use ROW() as well? Because duckdb's FILTER clause does not support
-    # a windowing context as of now: https://duckdb.org/docs/sql/query_syntax/filter.html
-    if (length(cols) > 1) {
-      stop("n_distinct(): Only one vector argument is currently supported when `na.rm = TRUE`.", call. = FALSE)
-    }
-
-    return(sql(paste0("COUNT(DISTINCT ", cols[[1]], ")")))
+    return(sql(paste0("COUNT(DISTINCT ", str_struct, ") FILTER (", str_null_check, ")")))
   } else {
-    # https://duckdb.org/docs/sql/data_types/struct.html#creating-structs-with-the-row-function
-    str_struct <- paste0("row(", paste0(list(...), collapse = ", "), ")")
-
     return(sql(paste0("COUNT(DISTINCT ", str_struct, ")")))
   }
 }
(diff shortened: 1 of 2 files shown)


tests/testthat/test-backend-dbplyr__duckdb_connection.R
```diff
@@ -278,9 +278,9 @@ test_that("aggregators translated correctly", {
   expect_equal(translate(n_distinct(x), window = FALSE), sql(r"{COUNT(DISTINCT row(x))}"))
   expect_equal(translate(n_distinct(x), window = TRUE), sql(r"{COUNT(DISTINCT row(x)) OVER ()}"))
   expect_equal(translate(n_distinct(x), window = TRUE, vars_group = "y"), sql(r"{COUNT(DISTINCT row(x)) OVER (PARTITION BY y)}"))
-  expect_equal(translate(n_distinct(x, na.rm = TRUE), window = FALSE), sql(r"{COUNT(DISTINCT x)}"))
-  expect_equal(translate(n_distinct(x, na.rm = TRUE), window = TRUE), sql(r"{COUNT(DISTINCT x) OVER ()}"))
-  expect_equal(translate(n_distinct(x, na.rm = TRUE), window = TRUE, vars_group = "y"), sql(r"{COUNT(DISTINCT x) OVER (PARTITION BY y)}"))
+  expect_equal(translate(n_distinct(x, na.rm = TRUE), window = FALSE), sql(r"{COUNT(DISTINCT row(x)) FILTER (x IS NOT NULL)}"))
```
krlmlr, line 281: I wonder if we can keep this SQL unchanged in the very common case of just one column.

R/backend-dbplyr__duckdb_connection.R
```diff
@@ -77,21 +77,14 @@ duckdb_n_distinct <- function(..., na.rm = FALSE) {
   }
   check_dots_unnamed()
 
-  if (!identical(na.rm, FALSE)) {
-    cols <- list(...)
+  # https://duckdb.org/docs/sql/data_types/struct.html#creating-structs-with-the-row-function
+  str_struct <- paste0("row(", paste0(list(...), collapse = ", "), ")")
```
krlmlr, line 81: Is `paste0()` the correct idiom here? The other SQL generators use `sql_expr()` (for simple expressions) or `build_sql()` (for nonstandard constructs).

krlmlr (commented):
Thanks, great!

R/backend-dbplyr__duckdb_connection.R
lschneiderbauer: That's currently handled the same way `n_distinct()` is handled in (base) dbplyr:
`translate(n_distinct(inject(something), y, na.rm = TRUE), window = FALSE)` in itself will translate anything:
```sql
<SQL> COUNT(DISTINCT row(inject(something), y)) FILTER (inject(something) IS NOT NULL AND y IS NOT NULL)
```
(This is true despite the usage of `glue_sql2()`.)

However, `mutate()` itself checks if the expressions given map to existing columns, and only then passes them on to the SQL translator:
```R
df |> 
  mutate(
    z = n_distinct(inject(something), na.rm = TRUE)
  )
```
```
Error in `mutate()`:
ℹ In argument: `z = n_distinct(inject(something), na.rm = TRUE)`
Caused by error:
! Object `something` not found.
```

I cannot comment on whether it was a design-decision to not let `glue_sql2()` perform more checks or whether this is an oversight.

krlmlr:
Thanks!

PR #1614, opened 2025-10-06
feat : add support for wildcards in `tbl_file()` paths

rplsmn:
Remove check for file.exist
Add test for what happens when file doesn't exist or invalid

I'm not sure about the test. I'd like it to cover the case where the wildcard points to a folder with incompatible schemas parquet files.


rplsmn:
I've played around with it : it actually loads the first file encountered among the valid wildcard paths. Even if there are multiple files with different schemas, DuckDB will simply load the first match.

I can remove the second test, the first one does work (wrong path).

krlmlr:
Thanks!

rplsmn:
Thank you for carrying me through this !

PR #1819, opened 2025-12-09
docs: Add alternative installation method to README

szarnyasg:


krlmlr:
Thanks, nice! I'll think about how to best incorporate this info, I see how it's especially important for DuckDB.

PR #1830, opened 2025-12-21
using cpp4r

pachadotdev:
Hi!

This follows from https://github.com/r-lib/cpp11/issues/401

I started cpp4r a while ago, motivated after the PR above and others less relevant for this project (such as support for complex numbers).

This is a draft PR. I kept the cpp11 headers to compare.

Best,
MVS


krlmlr:
Thank you for your contribution.

Currently, I'm comfortable using cpp11 with a patch stack of the necessary changes. This gives me the most flexibility and at the same time ensures correctness. The patch stack is here: https://github.com/r-lib/cpp11/compare/main...krlmlr:cpp11:main .

What issues do you think the move to cpp4r would solve?

pachadotdev:
> Thank you for your contribution.
> 
> Currently, I'm comfortable using cpp11 with a patch stack of the necessary changes. This gives me the most flexibility and at the same time ensures correctness. The patch stack is here: [r-lib/cpp11@main...krlmlr:cpp11:main](https://github.com/r-lib/cpp11/compare/main...krlmlr:cpp11:main) .
> 
> What issues do you think the move to cpp4r would solve?

that's good, if it works then it's good :)

cpp4r already has a very similar stack of changes and other things I have found while writing the PhD thesis

Issue #2308, opened 2026-04-02
Support VARIANT type


thohan88:
Thanks for all the work on maintaining this package.
It would be great if it could support the newly added [VARIANT type](https://duckdb.org/2026/03/09/announcing-duckdb-150#variant-type) introduced in 1.5.0.
See also: https://github.com/duckdb/duckdb/pull/18609

```R
library(duckdb)
#> Loading required package: DBI
con <- dbConnect(duckdb())
foo <- dbExecute(con, "CREATE TABLE events (id INTEGER, data VARIANT);")
foo <- dbExecute(con, "INSERT INTO events VALUES
    (1, 42::VARIANT),
    (2, 'hello world'::VARIANT),
    (3, [1, 2, 3]::VARIANT),
    (4, {'name': 'Alice', 'age': 30}::VARIANT);")

dbGetQuery(con, "FROM events;")
#> Error in `dbSendQuery()`:
#> ! Unknown column type for prepare: VARIANT
#> ℹ Context: rapi_prepare
```

<sup>Created on 2026-04-02 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">

<summary>

Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.5.0 (2025-04-11)
#>  os       Ubuntu 24.04.2 LTS
#>  system   x86_64, linux-gnu
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       Etc/UTC
#>  date     2026-04-02
#>  pandoc   3.6.4 @ /usr/bin/ (via rmarkdown)
#>  quarto   1.5.57 @ /usr/local/bin/quarto
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date (UTC) lib source
#>  cli           3.6.4   2025-02-13 [1] RSPM (R 4.5.0)
#>  DBI         * 1.2.3   2024-06-02 [1] RSPM (R 4.5.0)
#>  digest        0.6.37  2024-08-19 [1] RSPM (R 4.5.0)
#>  duckdb      * 1.5.1   2026-03-26 [1] RSPM (R 4.5.0)
#>  evaluate      1.0.3   2025-01-10 [1] RSPM (R 4.5.0)
[...]

PR #2313, opened 2026-04-13
feat: Add native VARIANT data type support

thohan88:
Adds R-side glue code to decode DuckDB's `VARIANT` columns (`>=1.5.0`) into native R list objects instead of erroring on fetch. I have tried to keep this minimal to mainly address #2308 for supporting reads as I suspect R → VARIANT conversion is a different beast.
 
## What this does

- Registers `VARIANT` as a known type across the R type system (`VECSXP` / `"list"`)
- Decodes each VARIANT row via `VariantUtils::ConvertVariantToValue` → `ValueToSexp`, producing native R objects:
  - Primitives → scalar R values (`logical`, `integer`, `double`, `character`)
  - Objects → named R lists
  - Arrays → unnamed R lists
  - NULL → `NULL`
  - Timestamps/dates inside variants → `POSIXct`/`Date` with proper attributes
- Extends `ValueToSexp` to handle `LIST`, `MAP`, `STRUCT`, and `UNION` value types so that nested VARIANT structures recursively produce the correct R representation
- Changes `ValueToSexp` signature to `const Value&` / `const string&` for const-correctness
- Uses `convert_opts.timezone_out` for timezone-aware conversion of timestamps inside variants

## What this does NOT do

- **No vectorized fast-path**: Each row materializes an intermediate `Value` object before converting to SEXP (`Variant binary → Value → SEXP`). This works correctly but creates extra allocations for large columns. A future improvement could use DuckDB's `VariantVisitor<T>` template to go directly from variant binary to SEXP, skipping the intermediate `Value` tree.
- **No R → VARIANT write support**: R objects cannot be written back as VARIANT columns via `dbWriteTable` or parameterized queries. Users can work around this by inserting via SQL with explicit `::VARIANT` casts. *(Mapping arbitrary R lists to DuckDB VARIANT is non-trivial.)*

## Files changed (5 files, +195 / −2)

| File | Change |
|------|--------|
| `src/transform.cpp` | Add `VARIANT` case to `duckdb_r_typeof`, `duckdb_r_decorate`, and `duckdb_r_transform` |
| `src/types.cpp` | Register `VARIANT` → `"list"` in `DetectLogicalType` |
| `src/utils.cpp` | Add `LIST`/`MAP`/`STRUCT`/`UNION` cases to `ValueToSexp`; `const` signature |
| `src/include/typesr.hpp` | Update `ValueToSexp` declaration to `const` |
| `tests/testthat/test-variant.R` | 8 test cases (37 expectations): primitives, empty containers, deep nesting, timestamps, dates, batched fetch, heterogeneous objects, `LIST(VARIANT)` composability |

## Testing

- `testthat::test_local(filter = "^variant$")` — 37/37 pass
- Full suite — 1265 pass, 0 new failures

## Note

Implemented with the help of Claude Opus.


src/utils.cpp
```diff
@@ -325,6 +325,34 @@ SEXP RApiTypes::ValueToSexp(Value &val, string &timezone_config) {
 		return res;
 	}
 
+	case LogicalTypeId::LIST:
+	case LogicalTypeId::MAP: {
```
krlmlr: What's the purpose of this code? Where is it tested? Should it be a separate PR?

krlmlr (commented):
Thanks. Can we now unlock the variant type in the `test_all_types()` test?

src/utils.cpp
thohan88: > What's the purpose of this code?

The `ValueToSexp` function converts a DuckDB Value to an R SEXP. Before this PR, it only handled scalar types (BOOLEAN, INTEGER, DOUBLE, VARCHAR, TIMESTAMP, etc.) and fell through to throw `NotImplementedException` for compound types like LIST, MAP, STRUCT, and UNION.

These cases are needed because the VARIANT transform path works like this:
1. In [transform.cpp:672](https://github.com/thohan88/duckdb-r/blob/8d11872b65be6714116eb693019977e3dd7a1c89/src/transform.cpp#L672) `VariantUtils::ConvertVariantToValue(...)` decodes the shredded binary into a Value
2. That Value is then passed to `ValueToSexp(val, ...)` to convert it to R

A VARIANT can contain any DuckDB type, including lists, structs, maps, and unions. Without these new cases, decoding `{'a': 1, 'b': [2, 3]}::VARIANT` would crash at `ValueToSexp` when it hits the struct Value containing a nested list Value.

> Where is it tested:

By several tests in [test-variant.R](https://github.com/thohan88/duckdb-r/blob/8d11872b65be6714116eb693019977e3dd7a1c89/tests/testthat/test-variant.R):
- Line 13: `{'a': 1, 'b': [2, 3]}::VARIANT`: exercises STRUCT → ValueToSexp → LIST → ValueToSexp recursion
- Line 14: `[4, 5, 6]::VARIANT`:  exercises LIST case
- Line 49: `list_value()::VARIANT`:  empty list
- Line 54: `MAP()::VARIANT`: MAP case
- Line 62: deeply nested struct/list
- Line 123: heterogeneous struct

> Should it be a separate PR?

Maybe. These cases only exist to support the VARIANT feature. Before VARIANT, `ValueToSexp` was only called from [register.cpp:168](https://github.com/thohan88/duckdb-r/blob/8d11872b65be6714116eb693019977e3dd7a1c89/src/register.cpp#L168) for Arrow filter pushdown with scalar constant values (where LIST, MAP and STRUCT constants are extremely unlikely). IMO, they are probably not independently useful enough to justify a standalone PR. Splitting them out would mean this PR would not compile on its own, but I am happy to make a stacked PR if you prefer that.

thohan88:
> Thanks. Can we now unlock the variant type in the `test_all_types()` test?

I do not see any entries for variant columns in the vendored [test_all_types.cpp](https://github.com/thohan88/duckdb-r/blob/39333ab207c8191b4cab5e8fc1defbcaa02503f2/src/duckdb/src/function/table/system/test_all_types.cpp) yet. Once it is added upstream, I would assume it would work out of the box with this transform logic.

src/utils.cpp
krlmlr: IIRC, we already convert structs and maps from DuckDB to R. I wonder if we could reuse that code path instead, or somehow combine these code paths. (Without looking at it too closely, should the "struct" case return a data frame instead of a bare list?)

Is this an implication of the first non-goal? If the advantage of going straight from binary to data is code reuse, I'd prefer that.

thohan88: The first non-goal is about avoiding the intermediate `Value` allocation by using `VariantVisitor<T>` to go straight from binary to SEXP to improve performance.

It would not enable reuse of the existing transform.cpp code though, because the two paths operate on different data representations: 
- Existing LIST, MAP, STRUCT code in transform.cpp works on columnar `Vector` objects (e.g. `ListVector::GetData`, `StructVector::GetEntries`),
- Upstream `VariantUtils::ConvertVariantToValue` returns a scalar `Value` per row. 

Reusing `duckdb_r_transform` from a `Value` would mean wrapping each one in a single-element `Vector` and going through the full allocate/decorate/transform pipeline per cell, which is doable but adds per-cell overhead compared to the direct recursive conversion.

Good catch that the struct case should return a data frame (not a bare list) and the map case should return a key/value data frame matching existing behavior in transform.cpp. I can update `ValueToSexp` to add the proper `duckdb_r_df_decorate` calls for consistency, but fully understand if you would like to explore a different direction for code reuse.

krlmlr: I see how code reuse is hard here, and perhaps not even attractive. Does the following property always hold: a column returned from DuckDB wrapped into a list (element-wise) gives the same result as coercing that column to a variant and returning from DuckDB? What are the exceptions?

We could then use `test_all_types()` to check for this property, this ensures consistency moving forward.

thohan88: After attempting a couple of alternatives, I think reusing `duckdb_r_transform` is a viable option. 

I have refactored `ValueToSexp` to wrap individual values in a single length `Vector` and funnel them through the existing transformation pipeline (`duckdb_r_allocate` -> `duckdb_r_decorate` -> `duckdb_r_transform`). Doing so achieves structural parity, code reuse and respect for connection-level settings (e.g. `bigint` and `timezone_out`). We also get rid of the existing manual type switch logic in `utils.cpp` which was only used for arrow filter pushdown.

It introduces a small per-cell overhead by wrapping values in a single-element `Vector`, but I believe the gains make it worth the trade-off. I do not think scalar conversion logic will be the bottleneck for `VARIANT` data anyway.

All tests passed and confirmed types match when round-tripped through `VARIANT`.

thohan88:
I have updated the branch and squashed the history to suggest a cleaner implementation which ensures type parity and makes use of existing code. Polite disclaimer that C++ is not my primary language, but I have attempted to follow the existing style with some assistance from Claude Opus. 

See comment: https://github.com/duckdb/duckdb-r/pull/2313#discussion_r3082011030

>After attempting a couple of alternatives, I think reusing `duckdb_r_transform` is a viable option. 
>
>I have refactored `ValueToSexp` to wrap individual values in a single length `Vector` and funnel them through the existing transformation pipeline (`duckdb_r_allocate` -> `duckdb_r_decorate` -> `duckdb_r_transform`). Doing so achieves structural parity, code reuse and respect for connection-level settings (e.g. `bigint` and `timezone_out`). We also get rid of the existing manual type switch logic in `utils.cpp` which was only used for arrow filter pushdown.
>
>It introduces a small per-cell overhead by wrapping values in a single-element `Vector`, but I believe the gains make it worth the trade-off. I do not think scalar conversion logic will be the bottleneck for `VARIANT` data anyway.
>
>All tests passed and confirmed types match when round-tripped through `VARIANT`.

src/utils.cpp
```diff
@@ -272,61 +272,49 @@ Value RApiTypes::SexpToValue(SEXP valsexp, R_len_t idx, bool typed_logical_null)
...
+	vec.SetValue(0, val);
+
+	SEXP dest = duckdb_r_allocate(type, 1, "variant", opts, "ValueToSexp");
+	duckdb_r_decorate(type, dest, opts);
+	duckdb_r_transform(vec, dest, 0, 1, opts, "variant");
+
+	// Scalar TIMESTAMP (no TZ) must have tzone="" for Arrow pushdown compatibility
+	if (type.id() == LogicalTypeId::TIMESTAMP && TYPEOF(dest) == REALSXP) {
+		Rf_setAttrib(dest, RStrings::get().tzone_sym, StringsToSexp({""}));
 	}
```
krlmlr: What fails if we omit this?

src/utils.cpp
```diff
@@ -272,61 +272,49 @@ Value RApiTypes::SexpToValue(SEXP valsexp, R_len_t idx, bool typed_logical_null)
...
+	if (type.id() == LogicalTypeId::UNION) {
+		auto &val_ref = UnionValue::GetValue(val);
+		return ValueToSexp(val_ref, convert_opts);
+	}
+
+	// Enable array conversion for values inside VARIANT
+	ConvertOpts opts = convert_opts;
+	if (type.id() == LogicalTypeId::ARRAY) {
+		opts.array = ConvertOpts::ArrayConversion::MATRIX;
+	}
```
krlmlr: I'd prefer to respect user preferences here. Would removing this code (and just using `convert_opts`) mean that matrices can be returned inside variants with the appropriate `dbConnect()` options?

krlmlr (commented):
Thank you! This looks like a useful addition and a great simplification of the code. I have a few questions.

Just a heads-up, the 1.5.3 release is scheduled for Monday.

src/utils.cpp
thohan88: Agree! I have removed the hardcoded `MATRIX` flag in `ValueToSexp`. It now correctly respects the array option passed to `dbConnect()`. I have updated tests in `test-variant.R` to verify that arrays inside variants are handled according to user preference.

src/utils.cpp
thohan88: Omitting this causes `test-register_arrow.R` to fail due to Arrows type strictness. Without this line, DuckDB's plain `TIMESTAMP` (no timezone) is assigned a default `UTC` timezone by R/Arrow. This causes a crash when DuckDB later tries to filter this data against other no-timezone timestamp literals, as Arrow refuses to compare a 'Timestamp-with-TZ' to a 'Timestamp-without-TZ.' I have kept the original logic and confirmed it passes. Happy to add a more detailed code comment if you think it is warranted.

krlmlr:
Thanks for driving this. The plan is to merge #2333 first, which isolates some of the refactorings applied here, then merge this, then submit to CRAN.

PR #2327, opened 2026-05-20
feat: Store downloaded extensions inside the duckdb package install directory

krlmlr:
## Summary

Reported by Prof. Brian Ripley on the `duckspatial` CRAN check (May 2026): the shared cache at `~/.local/share/R/duckdb/extensions/v<ver>/<platform>/` is loaded by every R installation on the host. When those installations were built against different C++ toolchains, the loader picks up a binary built for the wrong ABI.

This PR moves the extension cache from `tools::R_user_dir("duckdb", "data")` into the duckdb package's installed library directory:

```
<system.file(package = "duckdb")>/extensions/v<version>/<duckdb_platform>/<ext>.duckdb_extension
```

Co-locating the cache with the package install guarantees that downloaded extensions are paired with the exact toolchain that built duckdb itself — regardless of C++ standard library, compiler version, or any other ABI-relevant setting. On reinstall (upgrade, downgrade, fresh install on a new R), the install directory is replaced and the cache is wiped together with the old binaries; the next download picks up extensions matching the new build.

Secrets continue to live under `R_user_dir()` so they survive package upgrades.

## Changes

- `R/extensions.R` — `default_extension_directory()` returns `system_file_path("extensions")`. `cleanup_user_directory()` drops `"extensions"` from its preserve list so legacy `R_user_dir/extensions/` caches from earlier versions are cleaned up on CRAN test teardown.
- `R/package.R` — adds `system_file_path()` helper that prefixes a path with `system.file(package = ...)`.
- `R/Driver.R` — comment block reflects the new layout.
- `tests/testthat/test-extension_path.R` — asserts the install-dir-based layout via `system.file(package = "duckdb")`.
- `.gitignore` / `.Rbuildignore` — ignore `/extensions/` so that `pkgload::load_all()` (which resolves `system.file()` to the source tree) doesn't pollute the working tree or `R CMD build` tarballs.

## What this does not fix

DuckDB upstream publishes one binary per platform string (`linux_amd64`, `osx_arm64`, `windows_amd64_mingw`, …) — the string does not encode libstdc++ vs libc++ or compiler version. So if duckdb-r is built against a toolchain DuckDB doesn't publish for (e.g. clang + libc++ on Linux), extension loading still fails the same way it did before. Resolving that would require upstream changes (per-runtime binaries, ABI-aware refusal in `LOAD`, or build-from-source).

What this PR closes is the *interference* half of the complaint: two R installs on one machine now have completely separate caches, so one install's downloads can no longer claim the slot another install would have used.

https://claude.ai/code/session_01TJRriMUFX9tHFzTrPo3vj1


krlmlr:
@cidree: Can you please take a look? Planning to release today. Thanks!

Cidree:
Hei @krlmlr !
I'm sorry that I won't have time to have a deep look at it.
I had a glimpse at the changes and it looks fine, but I won't be able to test it. So from my side, go on with it!

krlmlr:
Thank you for the response, crossing fingers.

PR #2345, opened 2026-05-21
chore: Add opt-in system libduckdb linking for faster development builds

krlmlr:
## Summary

This PR adds an opt-in mode to link the R package against a system-installed libduckdb instead of compiling ~1700 vendored C++ files. This reduces clean install time from 10–15 minutes to ~5 seconds, enabling faster iteration during development and in CI/CD smoke tests.

## Key Changes

- **New script `scripts/install-libduckdb.sh`**: Downloads and installs prebuilt libduckdb binaries matching the vendored DuckDB version. Supports Linux (x86_64, aarch64) and macOS (universal). Includes version detection from vendored sources and validation that installed library matches vendored commit.

- **Enhanced `configure` script**: Detects `DUCKDB_R_USE_SYSTEM_LIB` environment variable and implements system libduckdb linking path:
  - Resolves libduckdb location via `DUCKDB_R_LIB_DIR`, pkg-config, or standard system paths
  - Validates that system libduckdb matches vendored DuckDB commit using `DUCKDB_SOURCE_ID` string matching
  - Generates `Makevars.duckdb` to skip compilation of vendored sources and link against system library with rpath
  - Provides clear error messages if library not found or version mismatch detected

- **New `src/include/system-lib.mk`**: Makefile fragment that overrides `SOURCES` to empty when using system libduckdb, preventing compilation of vendored code.

- **CI/CD integration** (`.github/workflows/custom/before-install/action.yml`):
  - Automatically enables system libduckdb mode for smoke jobs on all repos except `krlmlr/duckdb-r` (which hosts the vendoring pipeline)
  - Runs `install-libduckdb.sh` before package installation

- **Matrix workflow update** (`.github/workflows/versions-matrix/action.R`): Adds explicit source-build row to test matrix on non-vendoring repos to ensure vendored compile path remains covered.

- **Documentation** (AGENTS.md): Added section documenting the fast build workflow with usage examples and caveats.

## Implementation Details

- **Version matching**: The configure script extracts `DUCKDB_VERSION` and `DUCKDB_SOURCE_ID` from vendored `pragma_version.cpp` and validates the system libduckdb contains the same `DUCKDB_SOURCE_ID` string using the `strings` utility. This ensures ABI-level compatibility since the glue code uses internal C++ headers.

- **Platform detection**: Handles Linux (x86_64, aarch64) and macOS (universal) with appropriate library extensions (.so vs .dylib).

- **Rpath linking**: System libduckdb is linked with `-Wl,-rpath` to ensure the installed package finds the library at runtime without requiring it in standard system paths.

- **Opt-in only**: The feature only applies to `R CMD INSTALL .` (development builds), not `R CMD build` (CRAN-style tarballs), since the resulting installation depends on libduckdb being present at runtime.

- **Graceful fallback**: If `DUCKDB_R_USE_SYSTEM_LIB` is not set, the build proceeds with the existing vendored source compilation path.

https://claude.ai/code/session_01YRZKuhkTUGFsws3JsKBupq


.github/workflows/custom/before-install/action.yml
```diff
@@ -7,3 +7,23 @@ runs:
...
+      if: github.repository != 'krlmlr/duckdb-r' && runner.os != 'Windows'
+      run: |
+        echo 'DUCKDB_R_USE_SYSTEM_LIB=1' | tee -a $GITHUB_ENV
+      shell: bash
+
+    - name: Install libduckdb (system-lib mode)
+      if: env.DUCKDB_R_USE_SYSTEM_LIB == '1' && runner.os != 'Windows'
+      run: |
+        if [ "${RUNNER_OS}" = "Linux" ]; then
+          sudo scripts/install-libduckdb.sh --prefix /usr/local
```
krlmlr: Call `sudo` as needed in the script on Linux, avoid special-casing here.

configure
```diff
@@ -15,13 +15,127 @@ else
...
+    fi
+  fi
+
+  # Write a Makevars.duckdb that skips compilation of vendored sources.
+  cat include/system-lib.mk > Makevars.duckdb
+  {
+    echo ""
+    echo "# Populated by configure (DUCKDB_R_USE_SYSTEM_LIB=1)"
+    echo "PKG_LIBS=-L${duckdb_lib_dir} -lduckdb -Wl,-rpath,${duckdb_lib_dir}"
+  } >> Makevars.duckdb
```
krlmlr: Does this change files that are otherwise committed to version control? Prefer writing to an ignored file that is included only if it exists.

krlmlr:
Merging right after release.

PR #1586, opened 2025-09-22
test: ensure be able to install duckdb extensions on release version

eitsupi:
Related to #1517

PR #2353, opened 2026-06-12
test: Add CRAN guards to prevent heavy C++ engine tests on CRAN

krlmlr:
## Summary
This PR implements a guard system to prevent DuckDB's heavy C++ engine from being exercised during CRAN checks, while ensuring the full test suite runs in the project's own CI. The solution uses an environment variable (`DUCKDB_R_RUN_TESTS`) to gate all C++ code access and includes an "engine poisoning" mechanism to verify the guards are effective.

## Key Changes

- **New CRAN guard module** (`R/cran-guard.R`): Provides `duckdb_run_tests_enabled()` and `examples_enabled()` functions that check the `DUCKDB_R_RUN_TESTS` environment variable.

- **Test suite gating** (`tests/testthat.R`): Runs the suite only when `DUCKDB_R_RUN_TESTS` is set, otherwise emits a verbose message explaining the intentional skip and how to opt in. The check is inlined directly in the file so it is self-contained.

- **Example gating**: All runnable examples that touch the engine (`R/sql.R`, `R/Driver.R`, `R/dbConnect__duckdb_driver.R`, `R/register.R`) use `@examplesIf duckdb:::examples_enabled()`, which emits the same guidance when skipped. Man pages regenerated accordingly.

- **C++ engine tripwire** (`src/include/rapi.hpp`, `src/database.cpp`): A lightweight `DUCKDB_R_POISON_GUARD()` macro — a `cpp11::stop()` one-liner guarded by `#ifdef DUCKDB_R_POISON_ENGINE`, a no-op otherwise — placed at `rapi_startup`, the entry point of every DuckDB session. When built with `-DDUCKDB_R_POISON_ENGINE`, any test or example that reaches the engine aborts.

- **CI configuration (generic-first)**:
  - `.github/versions-matrix.R`: adds a matrix entry that carries the flag through a **generic `env` field** (`env = "DUCKDB_R_POISON_ENGINE=true"`) and leaves `DUCKDB_R_RUN_TESTS` unset.
  - `.github/workflows/R-CMD-check.yaml` stays generic: it only gains a generic forwarding of `matrix.env` to the custom `before-install` action, plus a generic `SKIP_UPDATE_SNAPSHOTS` escape hatch on the `update-snapshots` step. No poison-specific logic.
  - `.github/workflows/custom/before-install/action.yml` (custom action) picks up the flag: it applies the matrix `env`, and when `DUCKDB_R_POISON_ENGINE` is present it writes `CPPFLAGS += -DDUCKDB_R_POISON_ENGINE` to `~/.R/Makevars`, leaves `DUCKDB_R_RUN_TESTS` unset, and sets `SKIP_UPDATE_SNAPSHOTS=true` (the snapshot updater runs `testthat::test_local()`, bypassing `tests/testthat.R`). Normal CI runs opt in with `DUCKDB_R_RUN_TESTS=true`.

## Notable Implementation Details

- The guard accepts multiple truthy values: `"true"`, `"1"`, `"yes"`, `"on"` (case-insensitive).
- The tripwire sits at `rapi_startup`, which every DuckDB session must pass through, so no test or example can reach the engine without aborting.
- User-facing messages are verbose and actionable: they explain that the skip is intentional to avoid checking on CRAN, point to the extensive CI test suite, and show how to opt in locally.
- The solution is intentionally redundant (guards on both the R and C++ sides) for robustness.

https://claude.ai/code/session_01FsvpmBiT1sj31nSMaDGEXw

### DIFF
--- a/.github/versions-matrix.R
+++ b/.github/versions-matrix.R
@@ -1,2 +1,13 @@
-# Add even earlier Windows versions
-data.frame(os = "windows-latest", r = r_versions[4:6])
+list(
+  # Add even earlier Windows versions
+  data.frame(os = "windows-latest", r = r_versions[4:6]),
+  # Build with the DuckDB C++ engine poisoned (-DDUCKDB_R_POISON) and *without*
+  # DUCKDB_R_RUN_TESTS. Any test or example that reaches the C++ engine aborts,
+  # so this run verifies that the CRAN guards keep the engine untouched.
+  data.frame(
+    os = "ubuntu-24.04",
+    r = r_versions[[2]],
+    poison = "true",
+    desc = "with poisoning, without DUCKDB_R_RUN_TESTS"
+  )
+)
--- a/.github/versions-matrix.R
+++ b/.github/versions-matrix.R
@@ -1,13 +1,15 @@
 list(
   # Add even earlier Windows versions
   data.frame(os = "windows-latest", r = r_versions[4:6]),
-  # Build with the DuckDB C++ engine poisoned (-DDUCKDB_R_POISON) and *without*
-  # DUCKDB_R_RUN_TESTS. Any test or example that reaches the C++ engine aborts,
-  # so this run verifies that the CRAN guards keep the engine untouched.
+  # Build the DuckDB C++ engine with a tripwire (-DDUCKDB_R_POISON_ENGINE) and
+  # *without* DUCKDB_R_RUN_TESTS. The flag travels through the generic "env"
+  # matrix field and is picked up by the custom before-install action. Any test
+  # or example that reaches the C++ engine then aborts, so this run verifies
+  # that the CRAN guards keep the engine untouched.
   data.frame(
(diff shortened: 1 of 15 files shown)


eitsupi (commented):
If you do this, no tests will be run on R-universe.
Because even checks like #1586 cease to work, there is a risk of corrupted binaries being distributed.

(I'm not sure if it's ideal to avoid all testing on CRAN in this way in the first place), but shouldn't tests at least be run on R-universe?

krlmlr:
Thanks, fair point. Working on reestablishing r-universe tests (by unconditionally enabling tests on GitHub Actions, I suspect this will be enough, but we'll see).

v1.5.3 didn't make it to CRAN because DuckDB check on their testing rig timed out. (We don't even know, today, if v1.5.3 would have passed these checks.) I'd argue it's not CRAN's job to test DuckDB, and we do a lot of checking here already.

krlmlr:
Now in #2358.

PR #2372, opened 2026-06-25
docs: CRAN-safe storage-location policy + behavior-neutral scaffolding (Phase 0)

krlmlr:
## Summary

Documents the storage-location policy for the duckdb R package and lands the
behavior-neutral groundwork for it. **No runtime behavior changes** vs `main`:
the resolved extension/secret directories, `home_directory`, and connection
messages are unchanged. The actual policy (tempdir-by-default, markers, the
`duckdb_*_storage()` functions) is staged for follow-ups per the plan.

## Background

CRAN flagged reverse dependencies **`dbSpatial`** and **`duckspatial`** (Kurt
Hornik, 2026-06-23): during checks the user library is remounted **read-only**,
and duckdb writes downloaded extensions *into the package install directory*
(`system.file(package="duckdb")/extensions/`, since #2327), producing
`"... Read-only file system"`. That violates the CRAN Policy ("packages should
not write … anywhere else on the file system apart from the R session's
temporary directory"). Hornik asked that a CRAN-safe location policy be
standardized in duckdb-r so revdeps don't each reinvent it.

## What's in this PR

**Documentation (the spec):**
- `?duckdb_storage` — catalogues every on-disk location DuckDB uses (extensions,
  secrets, temp/spill, home directory, logs, database files), the resolution
  chain, the marker mechanism, and the per-location reference. CRAN/design
  rationale is kept in plain comments, out of the help page.
- `?duckdb_storage_config` — the planned `duckdb_extension_storage()`,
  `duckdb_secret_storage()`, and `duckdb_storage_status()` functions.
- `plan/PLAN-storage-locations.md` — phased implementation roadmap with
  per-bullet tests and a final checklist.

**Phase 0 — behavior-neutral seams** (so the policy is testable without touching
the real filesystem, HOME, or package library):
- `package_install_dir()` (wraps `system.file()`; `system_file_path()` builds on
  it), `duckdb_shared_home()` (replaces the hard-coded `"~/.duckdb"` literal),
  `session_temp_dir()`, `dir_is_writable()` (a try-write, not `file.access`).
- `has_rlang()` caches `requireNamespace("rlang")` once per session (used in
  `.onLoad`); `check_dots_empty()` forwards to `rlang::check_dots_empty0()` or a
  base fallback.
- `tests/testthat/test-storage-seams.R` pins current behavior and confirms each
  seam is mockable via `local_mocked_bindings()`.

## Behavior

Verified behavior-neutral: a fresh connection still resolves
[...]

PR #2375, opened 2026-06-26
feat: Configurable storage locations for extensions and secrets

### LINKED ISSUE #2370: Make extensions directory configurable by envvar
#2327 moved the extensions directory to use `default_extension_directory()` -- `system_file_path("extensions")` -- but the package directory is often not user-writable, for example on Posit Connect.

Additionally, there are other situations where the `config` parameter is not accessible to users, e.g. in situations where the duckdb connection is created an owned by another package, e.g. in a Shiny app with a chatbot that uses ragnar the Shiny developer doesn't have access to the `config` parameter to change the extensions directory.

It'd be very helpful if duckdb could inspect and use an environment variable like `DUCKDB_EXTENSIONS_PATH`, falling back to the package installation path. Or if the package installation path isn't writable, falling back to a temp directory would also be helpful.

krlmlr:
Closes #2370.

Implements the storage-location policy documented in `?duckdb_storage` /
`?duckdb_storage_config` (merged in #2372). With no configuration, nothing is
written outside the R session's temporary directory; persistence is opt-in.

## What changes

By default, `duckdb()` resolves the engine's writable locations as follows:

| Location | Default | Setting |
|----------|---------|---------|
| Extension cache | package library when writable, else `tempdir()` | `extension_directory` |
| Stored secrets | `tempdir()` sub-directory (was `R_user_dir()`) | `secret_directory` |
| Temp/spill (in-memory DBs) | `tempdir()` sub-directory (was cwd `.tmp`) | `temp_directory` |
| Home | left untouched | `home_directory` |

Each is overridable via `config`, an `options(duckdb.<kind>_directory=)`, or a
`DUCKDB_<KIND>_DIRECTORY` env var. `home_directory` is deliberately left alone
so `~` in user SQL keeps its meaning.

A marker file (`.duckdb-r-keep`) opts a kind into a persistent root. The
write-probe of the package library doubles as its writability test, so on the
CRAN check farm (read-only library) the cache falls back to `tempdir()` with no
failed write. The first time the library cache is initialized, the package says
so once (effectively once per install).

## New user-facing functions

- `duckdb_extension_storage(location = c("session", "user", "shared", "library"), ..., migrate = TRUE, conflict = "error")`
- `duckdb_secret_storage(location = c("session", "user", "shared"), ..., migrate = TRUE, conflict = "error")`
- `duckdb_storage_status()` — prints a readable summary and invisibly returns a
  data frame (`kind`, `source`, `directory`).

`location` names a root (validated with `rlang::arg_match()`); `"session"` is
the default and the opt-out. Arbitrary paths go through the option/env var, not
these functions. `migrate` relocates the cached files (preserving DuckDB's
`v<version>/<platform>/` layout, sweeping every marked root) and `conflict`
(`"error"`/`"ours"`/`"theirs"`) decides name collisions.

## Removed

`duckdb_consolidate_secrets()` and the two-location startup message are gone (no
deprecation ceremony — they shipped only briefly); secret migration is now the
`migrate` step of `duckdb_secret_storage()`.

The reactive "your extensions/secrets won't persist" warning was prototyped here
but **removed and deferred to a separate change** — detecting the write and
picking a reliable, non-fragile trigger needs more thought (see the plan).

## Implementation notes

- Every dependency on the environment goes through a mockable seam
  (`session_temp_dir`, `default_user_directory`, `duckdb_shared_home`,
  `system_file_path`, …), so resolution/marker/migration logic is tested in
  pure R without touching the real filesystem or `$HOME`.
- `rlang` stays a soft dependency: `check_dots_empty0()`, `inform()`, and
  `arg_match()` have base fallbacks swapped for the rlang versions in `.onLoad()`.
- End-to-end tests (`skip_on_cran()`) create a persistent secret and install the
  `httpfs` extension against the real engine; the extension test skips
  best-effort when the download/install fails (no network).

https://claude.ai/code/session_01R5F1rPyR7Vo3r9BRqUXCfu


R/storage-locations.R
```diff
@@ -0,0 +1,396 @@
...
+
+# --- Marker files -------------------------------------------------------------
+
+# Name and contents of the per-kind "keep" marker. Contents are informational
+# only: the package checks for the file's presence, never reads it back.
+KEEP_MARKER_NAME <- ".duckdb-r-keep"
+KEEP_MARKER_TEXT <- paste(
+  "This directory is used by the duckdb R package to store downloaded",
+  "extensions or secrets. It is safe to delete."
+)
```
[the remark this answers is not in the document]
krlmlr: Agreed — the "It" was ambiguous and read as the directory. Reworded in 84b85de so the marker makes clear that deleting the *file* opts out, while deleting the directory removes the stored data.

---

R/storage-locations.R
```diff
@@ -0,0 +1,396 @@
...
+}
+
+# --- Ephemeral-storage message ------------------------------------------------
+
+# TRUE if `path` lies inside the session's temporary directory.
+path_within_tempdir <- function(path) {
+  np <- normalizePath(path, winslash = "/", mustWork = FALSE)
+  tp <- normalizePath(session_temp_dir(), winslash = "/", mustWork = FALSE)
+  startsWith(np, tp)
+}
```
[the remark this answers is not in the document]
krlmlr: Fixed in 84b85de: it now compares whole path segments (`np == tp || startsWith(np, paste0(tp, "/"))`), so a sibling like `/tmp/Rtmp1234` no longer matches `/tmp/Rtmp123`, and it returns `FALSE` for empty/NA/non-scalar input.

---

R/storage-locations.R
```diff
@@ -0,0 +1,396 @@
...
+    if (
+      !file.copy(file.path(from, r), dst, overwrite = TRUE, copy.date = TRUE)
+    ) {
+      stop("Failed to migrate ", file.path(from, r), ".", call. = FALSE)
+    }
+  }
+  # Every source is removed: copied ones moved, "theirs" collisions dropped.
+  unlink(file.path(from, rel))
+  invisible()
+}
```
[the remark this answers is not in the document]
krlmlr: Fixed in 84b85de: after the copy, `migrate_storage()` re-checks which sources still exist and emits a `warning()` listing the files that could not be removed, so the user is told when data ends up duplicated across locations rather than it happening silently.

---

plan/PLAN-storage-locations.md
```diff
@@ -54,77 +54,77 @@ resolution/marker logic calls these, never the underlying primitive.
       same strategy as `is_interactive` / `rapi_error`; no rlang-availability
       cache).
 - [x] `is_interactive()` — pre-existing; use it, never `interactive()`.
-- [ ] `session_temp_dir()` (wraps `tempdir()`) — deferred to Phase 1.
-- [ ] Options/env reads stay `getOption()` / `Sys.getenv()` (mockable via
+- [x] `session_temp_dir()` (wraps `tempdir()`) — deferred to Phase 1.
```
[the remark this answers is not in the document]
krlmlr: Fixed in 84b85de — the bullet now reads "implemented in Phase 1".

---

R/storage-locations.R
```diff
@@ -0,0 +1,396 @@
...
+
+# The directory a (root, kind) pair maps to, e.g. `<root>/extensions`. An
+# explicit path (anything that is not a known root name) is used verbatim.
+storage_dir <- function(root, kind) {
+  known <- c("session", "user", "shared", "library")
+  if (!root %in% known) {
+    return(path.expand(root))
+  }
+  file.path(storage_root_base(root), storage_kind_subdir(kind))
+}
```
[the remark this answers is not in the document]
krlmlr: Good catch — that was a real inconsistency. Fixed in 8dbb4da: `duckdb_*_storage()` now rejects a non-root `location` (with an error pointing to `options(duckdb.<kind>_directory=)` / the env var for arbitrary directories), since a marker placed in an arbitrary path could never be rediscovered. `storage_dir()` is root-only accordingly, and the docs no longer claim an explicit path is accepted.

---

R/storage-config.R
```diff
@@ -62,32 +54,42 @@
 #'
 #' @seealso [duckdb_storage] for the storage policy these functions implement.
 #' @name duckdb_storage_config
-#' @aliases duckdb_extension_storage duckdb_secret_storage duckdb_storage_status
 #' @keywords internal
```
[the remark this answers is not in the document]
krlmlr: Agreed. Fixed in 8dbb4da: dropped `@keywords internal` from both `duckdb_storage_config` (the exported functions) and the `duckdb_storage` policy page, so the public API now shows in the help index. (They were already listed explicitly in `_pkgdown.yml`.)

---

R/Driver.R
```diff
@@ -66,19 +66,24 @@ duckdb <- function(
...
   if (!("secret_directory" %in% names(config))) {
     config["secret_directory"] <- resolve_secret_directory()
   }
+  if (!("temp_directory" %in% names(config))) {
+    temp_directory <- resolve_temp_directory(dbdir)
+    if (!is.null(temp_directory)) {
+      config["temp_directory"] <- temp_directory
+    }
+  }
+  maybe_ephemeral_state_message(config[["extension_directory"]])
```
krlmlr: Only if not provided by user?

krlmlr: Right. Fixed in bdd60a5: the message now fires only when we chose the location ourselves — i.e. `extension_directory` is not in `config` *and* there's no `options(duckdb.extension_directory=)` / `DUCKDB_EXTENSION_DIRECTORY` override. If the user set it any of those ways the choice is theirs, and the message (which points at exactly those knobs) would just be noise.

---

R/storage-config.R
```diff
@@ -38,18 +33,17 @@
...
 #' consent to write outside the temporary directory.
 #'
-#' `duckdb_secret_storage()` replaces `duckdb_consolidate_secrets()`: secret
-#' migration is now a `migrate` step of `duckdb_secret_storage()`.
-#'
-#' @param location The destination root, or an explicit path. Recognized roots
-#'   are `"session"` (the per-session temporary directory; also the opt-out --
-#'   it removes the marker), `"user"` ([tools::R_user_dir()]), `"shared"`
+#' @param location The destination root (not a path). Recognized roots are
+#'   `"session"` (the per-session temporary directory; also the opt-out -- it
```
krlmlr: Would a bullet list work here?

R/storage-config.R
```diff
@@ -58,36 +52,48 @@
...
+#' @rdname duckdb_storage_config
+#' @export
 duckdb_extension_storage <- function(
   location,
   ...,
   migrate = TRUE,
   conflict = "error"
 ) {
-  stop("`duckdb_extension_storage()` is not implemented yet.", call. = FALSE)
+  check_dots_empty(...)
```
krlmlr: If we use `check_dots_empty0()`, our shim should also be named like this.

R/Driver.R
krlmlr: Done in 5e2f519 — the four recognized roots are now an itemized list in the `location` param rather than a run-on sentence.

---

R/storage-config.R
```diff
@@ -58,36 +52,48 @@
...
-  stop("`duckdb_secret_storage()` is not implemented yet.", call. = FALSE)
+  check_dots_empty(...)
+  set_storage_marker("secrets", location, migrate, conflict)
 }
 
+#' @rdname duckdb_storage_config
+#' @export
 duckdb_storage_status <- function() {
-  stop("`duckdb_storage_status()` is not implemented yet.", call. = FALSE)
+  kinds <- c("extensions", "secrets")
```
krlmlr: Unroll loops.

R/storage-locations.R
```diff
@@ -0,0 +1,456 @@
...
+session_temp_dir <- function() {
+  tempdir()
+}
+
+# Per-kind sub-directory name under a storage root.
+storage_kind_subdir <- function(kind) {
+  switch(
+    kind,
+    extensions = "extensions",
+    secrets = "stored_secrets",
```
krlmlr: If DuckDB uses stored_secrets, we should use `"stored_secrets"` instead of `"secrets"` everywhere for consistency. Also, use "stored secrets" where appropriate in the docs.

R/storage-locations.R
```diff
@@ -0,0 +1,456 @@
...
+persistent_roots <- function() {
+  c("user", "shared")
+}
+
+# --- Marker files -------------------------------------------------------------
+
+# Name and contents of the per-kind "keep" marker. Contents are informational
+# only: the package checks for the file's presence, never reads it back.
+KEEP_MARKER_NAME <- ".duckdb-r-keep"
+KEEP_MARKER_TEXT <- paste(
```
krlmlr: `paste0()`

R/storage-config.R
krlmlr: Done in 5e2f519 — the four recognized roots are now an itemized bullet list in the `location` param. (Apologies, my acknowledgement for this landed on the Driver.R thread by mistake earlier.)

---

R/storage-config.R
krlmlr: Agreed — renamed in 07c3cf6. The base shim is now `check_dots_empty0()` (definition in `R/package.R`, the `.onLoad()` swap, and both call sites), matching the `rlang::check_dots_empty0()` it shadows.

---

R/storage-config.R
krlmlr: Done in 07c3cf6 — `duckdb_storage_status()` now calls `describe_storage()` explicitly for each of the two kinds instead of looping with `lapply`/`vapply`.

---

R/storage-locations.R
```diff
@@ -0,0 +1,456 @@
...
+}
+
+# Write the keep-marker into `dir` (creating `dir` if needed) and leave it in
+# place. The write doubles as a writability probe: returns TRUE if the marker
+# now exists. A real write rather than file.access(), which is unreliable on
+# Windows.
+write_keep_marker <- function(dir) {
+  if (
+    !dir.exists(dir) &&
+      !dir.create(dir, recursive = TRUE, showWarnings = FALSE)
```
krlmlr: Two nested `if()` are better here, store the output of `dir.create()` in a variable.

R/storage-locations.R
krlmlr: Done in 900042c. The internal storage-kind token is now `"stored_secrets"` everywhere (matching DuckDB's sub-directory), and the docs/messages say "stored secrets". Since the kind now equals the sub-directory name, `storage_kind_subdir()` collapsed into `storage_dir()` (which validates the kind). I kept the DuckDB setting / option / env / function names on `secret` (`secret_directory`, `duckdb.secret_directory`, `DUCKDB_SECRET_DIRECTORY`, `duckdb_secret_storage()`), since those are DuckDB's own names; let me know if you'd like those touched too.

---

R/storage-config.R
krlmlr: Also in 900042c: `KEEP_MARKER_TEXT` now uses `paste0()` with explicit spacing rather than relying on `paste()`'s separator.

---

R/storage-locations.R
krlmlr: Done in 900042c — `KEEP_MARKER_TEXT` now uses `paste0()` with explicit trailing spaces instead of relying on `paste()`'s separator. (My earlier note about this landed on the wrong thread — this is the right one.)

---

R/storage-locations.R
krlmlr: Done in 900042c — `write_keep_marker()` now nests the two checks and stores the `dir.create()` result:

```r
if (!dir.exists(dir)) {
  created <- dir.create(dir, recursive = TRUE, showWarnings = FALSE)
  if (!created) {
    return(FALSE)
  }
}
```

---

tests/testthat/test-secret_directory.R
```diff
@@ -1,251 +1,40 @@
...
-    withCallingHandlers(
-      maybe_secret_directory_message(),
-      packageStartupMessage = function(m) {
-        message(conditionMessage(m), appendLF = FALSE)
-        invokeRestart("muffleMessage")
-      }
-    ),
-    "stored secrets exist in two locations"
-  )
+  withr::local_options(duckdb.secret_directory = NULL)
```
krlmlr, line 10: A separate test block is cleaner here.

tests/testthat/test-storage-config.R
```diff
@@ -0,0 +1,193 @@
+# User-facing storage configuration (plan/PLAN-storage-locations.md, Phase 3).
+
+local_storage_roots <- function(.local_envir = parent.frame()) {
+  roots <- list(
+    user = withr::local_tempdir(.local_envir = .local_envir),
+    shared = withr::local_tempdir(.local_envir = .local_envir),
+    session = withr::local_tempdir(.local_envir = .local_envir),
+    library = withr::local_tempdir(.local_envir = .local_envir)
```
krlmlr: Use unambiguous names for the temp dirs.

tests/testthat/test-secret_directory.R
krlmlr: Done in d6f3f9b — split into two `test_that()` blocks, one for the option override and one for the env-var override.

---

tests/testthat/test-storage-config.R
krlmlr: Done in 9659e60 — each `withr::local_tempdir()` now takes a role prefix (`"user-"`, `"shared-"`, `"session-"`, `"library-"`), so the four roots are distinguishable in test output instead of all sharing the default `file` prefix.

---

PR #2437, opened 2026-07-30
feat(sync): Mirror the release branches, not only `main`

krlmlr:
Phase 3 of [`plan/PLAN-vendoring-simplification.md`](../blob/main/plan/PLAN-vendoring-simplification.md), pulled ahead of the fork move so the fix serves the standalone repo too.

## The defect

The `Flavors` table counts each `.dev` series *ahead* of the branch it releases from, and shields.io compares **within a single repository** — so the base of that comparison has to exist in `krlmlr/duckdb-r`, not only in `duckdb/duckdb-r`.

`sync.yaml` fast-forwarded `main` and nothing else.
That makes this a live defect and not only a fork-migration item: today's `duckdb.1.4.dev` *ahead* badge compares against a `v1.4-andium` mirror that nothing refreshes, so it counts commits that have already shipped.

## The change

The job mirrors `main` plus every release line it finds upstream — `v<major>.<minor>-<codename>`, and its `-lts` sibling — **discovered from the refs** rather than configured, the way a series is.
A branch the fork does not have yet is created, which also retires the manual "mirror the release branch into the fork" step in `series-open.md`.

Fast-forward only, never force.
A mirror carrying a commit upstream does not have is skipped and the job then fails: the other mirrors still move, and a mirror with a commit of its own is not a state an hourly job should sit on quietly.

Two mechanical notes:

* upstream is now fetched **directly** rather than through a clone. A clone keeps only its own default branch under `refs/heads/`, so refetching from it would have found `main` and nothing else — the release branches would have been silently skipped. (This is the bug the first draft had, caught in testing.)
* no `git-identity` step: nothing here creates a commit any more. Each branch moves by pushing a commit that already exists upstream, which is what keeps this a mirror rather than a merge.

## Verification

Simulated against a fake upstream and a fake fork covering all four states at once:

| branch | state | outcome |
|---|---|---|
| `main` | behind | fast-forwarded |
| `v1.4-andium` | behind | fast-forwarded |
| `v1.5-variegata` | missing in the fork | created |
| `v1.4-andium-lts` | carries a local commit | warned, skipped, untouched; job exits non-zero |
| `b-re2`, `gh-pages` | not release lines | never considered |

---
_Generated by [Claude Code](https://claude.ai/code/session_01RXpDkonMiTRwe147p6LBm8)_

### DIFF
--- a/.claude/skills/series-open.md
+++ b/.claude/skills/series-open.md
@@ -103,11 +103,15 @@ Two things to check before pushing:
   and the fork is disconnected from `duckdb/duckdb-r`,
   so a base branch that exists only in the canonical repo
   renders as an error, not a count.
-  Mirror the release branch into the fork —
-  and keep the mirror fresh:
-  `sync.yaml` fast-forwards only `main`,
-  so a mirror left behind makes *ahead*
-  count commits that have already shipped.
+  `sync.yaml` mirrors the release branches hourly
+  alongside `main`, creating a mirror that does not exist yet
+  and fast-forwarding one that does,
+  so the base of an *ahead* badge is there and stays fresh —
+  as long as the upstream branch is named like a release line
+  (`v<major>.<minor>-<codename>`, optionally `-lts`),
+  which is what that job discovers by.
+  Check the run once after a release cut;
+  a stale mirror makes *ahead* count commits that have already shipped.
 * **The table must stay clear of `scripts/flavor.patch`.**
   `README.md` is a flavored file;
   the patch rewrites the installation hunks near the top.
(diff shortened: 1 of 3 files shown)


krlmlr:
Won't merge, will be done by the pull app.

krlmlr:
Closing unmerged: the mirroring will come from the **Pull app** on a real fork instead.

That is the better mechanism, and it makes this workflow unnecessary rather than merely redundant — `sync.yaml` exists only because the standalone copy is not a fork and has no upstream to be pulled from. Replacing the copy replaces the need.

The defect this PR was written for is still real, so it stays recorded rather than disappearing with the branch: the `duckdb.1.4.dev` *ahead* badge compares against a `v1.4-andium` mirror nothing refreshes, and until the fork move the mirrors are refreshed by hand. #2450 updates §6 and §9 of the plan accordingly, and names this approach as considered-and-rejected so it does not get re-derived.

The branch stays on the remote if the argument is ever wanted again — the simulated four-state test (behind / missing / diverged / not-a-release-line) is in its description.

---

Issue #2306, opened 2026-03-31
Autoload of ICU extension does not work


al-obrien:
## Issue description

Both on version 1.4.2 and the latest 1.5.1, duckdb from an R session on RHEL8 does not appear to have the important `icu` extension loaded. 

Others (https://github.com/duckdb/duckdb-r/issues/1084) have referenced this previously but it appears to be non functional in this situation. 

<img width="1158" height="578" alt="Image" src="https://github.com/user-attachments/assets/92da543f-c33e-4dad-af98-31f6942d4c07" />  

These are the duckdb extension settings:

<img width="1301" height="165" alt="Image" src="https://github.com/user-attachments/assets/9d8a02ab-1fe7-42e4-81fd-145845724c6d" />

This is confusing as the official docs have a reference to this being bundled:  

<img width="789" height="665" alt="Image" src="https://github.com/user-attachments/assets/624fc971-7b16-4151-92bb-927757edc546" />

## Session Info

```
> sessionInfo()
R version 4.4.1 (2024-06-14)
Platform: x86_64-pc-linux-gnu
Running under: Red Hat Enterprise Linux 8.10 (Ootpa)

Matrix products: default
BLAS/LAPACK: /usr/lib64/libopenblasp-r0.3.15.so;  LAPACK version 3.9.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8   
 [6] LC_MESSAGES=en_US.UTF-8    LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

time zone: America/Edmonton
tzcode source: system (glibc)

attached base packages:
[1] stats     graphics  grDevices datasets  utils     methods   base     

other attached packages:
 [1] data.table_1.17.8  qs2_0.1.6          fs_1.6.6           AHRtools_1.2.4     AHRio_0.7.2        odbc_1.6.3         DBI_1.2.3         
 [8] stringr_1.6.0      glue_1.8.0         lubridate_1.9.4    crew_1.3.0         tarchetypes_0.13.2 targets_1.11.4    

loaded via a namespace (and not attached):
[...]

Issue #22, opened 2023-10-03
any tricks to speed up install time


jangorecki:
Any known tricks to speed up install time?
Also this that happens during R CMD check
```
* checking whether package 'duckdb' can be installed ...
 [35m/35m] OK
```
35 minutes is quite a blocker for comfortably testing changes...

Tmonster:
Are you talking about just in the GitHub actions? Or locally? Locally you can try to use CCACHE

https://dirk.eddelbuettel.com/blog/2017/11/27/

https://stackoverflow.com/q/45512586/946850

You can also enable parallel builds with 
`MAKE="make -j{ncores}" R CMD check .` or  `MAKE="make -j{ncores}" R CMD INSTALL .`

This has made my local development time much faster

jangorecki:
Sounds good, will test next time when installing. What seems reasonable is to include it somewhere on the readme, so users can find that straight away.

eitsupi:
Sorry if this is not possible as I am not familiar with the internal structure of this package, but I am wondering if it is possible to download and use the pre-compiled libduckdb [as the `arrow` package does (on Linux)](https://arrow.apache.org/docs/11.0/r/articles/install.html#r-source-package-with-libarrow-binary).

jangorecki:
Or at least reused compiled code when doing changes only R or Rd files

krlmlr:
We'll make `scripts/install-libduckdb.sh` more accessible (e.g., by not requiring root) for Linux and macOS users, will also investigate if it works on Windows. This will dramatically shorten the edit-compile loop (even without ccache). For installs of binary versions of the package, we'll ensure that the front page docs point to the right instructions or even contain them.

PR #2443, opened 2026-07-30
docs: Extend the documentation tree to the whole package, starting with `BUILD.md` and `?duckdb_extensions`

krlmlr:
## What this is

`plan/PLAN-vendoring-simplification.md` §8 designed a documentation tree for the vendoring work. This reviews that design, generalizes it to the whole package — semantics, build, tooling, tests, CI — and lands the first two nodes.

Prioritization comes from the triage in `plan/PLAN-inbox-zero.md` (branch `claude/review-open-issues-prs-46alwv` on `krlmlr/duckdb-r`): the documentation that lets triaged issues close is written first.

**Docs only.** The `configure` fix this work turned up is split out into #2446, which this PR does not depend on — `BUILD.md` describes the tree as it is today and points at that PR.

## The review

§8's four rules hold and are adopted unchanged. Five findings do not:

1. **The tree has two roots but only one subtree.** Every node §8 routes to is maintainer-facing, so a user following `README.md` lands in the branch model. That missing half is why eight issues triaged `CLOSE-DOCS` have nowhere to be answered.
2. **The medium is assumed to be Markdown in the repo.** For the user-facing half it has to be `.Rd` — that is what `?topic`, <https://r.duckdb.org/>, and the offline CRAN tarball serve. `_pkgdown.yml` is that subtree's routing table, a role §8 does not name.
3. **"No orphan docs" is the less damaging orphan problem.** Orphan *surface* — a script, workflow, or build knob no document owns — is found only when someone files an issue. §8 itself records two instances found by hand.
4. **Single-owner is asserted per topic, but topics are never enumerated.** At ~160 non-vendored, non-generated files the list has to run topics → owners, so *unowned* is computable rather than discovered.
5. **The tree is sequenced behind work it does not depend on** (Phase 4, after the vendoring kernel).

Plus one correction: §8 assigns the script inventory to `scripts/VENDORING.md`, but `scripts/` also holds build, flavor, and release tooling. **Inventories are owned by purpose, not by directory.**

## What lands here

**`plan/PLAN-docs-tree.md`** — two roots, two subtrees, and a fifth rule: every path has exactly one owning document, checkable via a manifest plus `scripts/docs-coverage.R` rather than discovered by review. Maps each of the five semantics pages to the issues it closes, and includes a section on why a generated wiki (DeepWiki-style) is a complement rather than the plan — it solves discovery, not ownership, and cannot recover knowledge that only exists in issue threads.

**`BUILD.md`** — the first engineering node. Owns `configure`, the makefile fragments, every build knob, the `.dd` files, the tarball layout, formatting, and the C++ warning policy. Absorbs those sections from `AGENTS.md`, which now routes instead of explaining. Answers #22.

**`?duckdb_extensions`** (`R/concepts-extensions.R`) — the first semantics leaf, and the answer to #2306. `_pkgdown.yml` gains the `Guides` section the remaining four pages land into.

## Why `BUILD.md` leads with what the fast path hides

Two ways a supposedly vendored build reports the *release* library's configuration, both documented under "Switching between modes":

- a stale `src/Makevars.system-lib` keeps emptying `SOURCES` and linking `-lduckdb` after the opt-in is unset (#2446 fixes this in `configure`);
- a stale `src/duckdb.so` survives the switch in the other direction — with `SOURCES` empty it is newer than the glue objects, so make relinks nothing and the fast path installs the vendored library. Not fixable in `configure`.

`ldd` on the installed shared object is given as the reliable test of which mode produced a package.

## Verification

Facts in the page are verified against a genuinely vendored build (confirmed by `ldd`: no `libduckdb` dependency):

| | Vendored build | Release libduckdb (fast path) |
|---|---|---|
| Statically linked | `core_functions`, `parquet` | also `icu`, `json`, `autocomplete` |
| `autoinstall_known_extensions` | `false` | `true` |
| `autoload_known_extensions` | `true` | `true` |

Both follow from `src/Makevars.in` defining `DUCKDB_EXTENSION_AUTOLOAD_DEFAULT` but not `DUCKDB_EXTENSION_AUTOINSTALL_DEFAULT`.

- Test suite: **FAIL 0 | WARN 0 | SKIP 60 | PASS 1265**.
- `scripts/flavor-package-name.R` caught hard-coded package names in an early draft of `BUILD.md`; resolved by pointing at `README.md` rather than duplicating the install instructions — which the single-owner rule wanted anyway. Scan is now clean.
- `roxygen2::roxygenize()` with the pinned `8.0.0.9000` writes only the new `.Rd`; `checkRd()` clean.

## Scope, honestly

This is phase A of five. The Markdown tree grows by exactly one file (`BUILD.md`); the plan's other four engineering nodes, four semantics pages, and the coverage checker are design only. Of the 8 `CLOSE-DOCS` issues, this closes one (#2306), plus #22 via `BUILD.md`.

Follow-ups, one PR each: `?duckdb_types` → `?duckdb_connections` → `?duckdb_csv` → `?duckdb_memory`, then `TESTING.md`, `src/GLUE.md`, `R/ARCHITECTURE.md`, `.github/CI.md`, then the coverage check.


krlmlr:
Superseded by #2454.

PR #2459, opened 2026-08-02
docs(handbook): Write the meta/plans leaf and generate `plan/README.md`

krlmlr:
## What this leaf now owns

`handbook/meta/plans/` owns everything the hand-written `plan/README.md` said, rewritten as leaf prose: what belongs under `plan/`, the plan-versus-history distinction and why the two live in separate directories, the caveat that a document there may have been overtaken and the owner of the topic today wins, and the conventions for adding, retiring, and not orphaning a document. It records as a decision — not as an accident of layout — that `plan/` stays outside the handbook: the handbook describes the package as it is today, a plan describes intent, and filing a proposal among the descriptions would make the tree assert things that are not so. That decision is exactly what makes the generated in-place index necessary, so the leaf also explains the index and the ownership map behind it. Its limits: it tracks no individual plan (what a plan proposes and how far along it is belongs to the plan) and it holds no inventory.

`plan/README.md` is no longer hand-written. It is generated by `docs-readme.R`, one row per document, grouped under the handbook leaf that owns each document's topic, and its intro backreferences this leaf.

## What moved

- `plan/README.md` — fully absorbed and replaced by the generated index. The prose (what the directory is for, the plan/history split, the overtaken-document caveat, "Adding a document here", the orphan rule) is now leaf prose; the two hand-maintained tables are now one generated table per owning leaf. The orphan rule became mechanical: an unmapped file renders under `Unowned` and `--check` fails.
- `.claude/skills/docs-consistency/docs-readme.R` — generalised from a `scripts/`-only generator into one that renders an index for both `scripts/` and `plan/`. The file list, the target path, the H1, the intro block, and the `groups` are now per-directory entries in a `directories` list; the ownership map is one map with a slice per directory; `--check` covers every directory it generates and reports each one. Files are listed by their path *relative to the directory* rather than by basename, which is what makes `plan/history/` render correctly (`history/vendoring-loop.md`); `scripts/` has no subdirectory, so its index regenerates byte-identically to `main`. The design is otherwise untouched: purpose extracted from each file's own header (Markdown H1 for plan documents), first-match-wins with the double-owned warning, groups sorted by handbook path.
- `scripts/README.md` — regenerated; byte-identical to `main` (`git diff origin/main -- scripts/README.md` is empty). Included only as proof the generalisation is behaviour-preserving.
- `.claude/skills/docs-consistency/SKILL.md` — the helper now renders two indexes, and check 4 (source-to-leaf coverage) names `plan/` alongside `scripts/` as machine-mapped.
- `README.md` (root) — the `plan/README.md` bullet in `## Documentation` now names `handbook/meta/plans/`, with `plan/` linked as the index.
- `AGENTS.md` — the "Designs, plans, and historical documents" row now routes to `handbook/meta/plans/`, with `plan/README.md` named as the directory index.
- `handbook/meta/handbook/README.md` — "The forms" said the generated index covers "`scripts/` alone" and that extending it was open work. It now names both `scripts/` and `plan/`, says they are slices of one file-to-leaf ownership map, and keeps the remaining directories as open work.

## The `plan/` ownership map

Each document is mapped to the leaf whose topic it carries. `README.md` maps to `meta/plans/`, mirroring `scripts/README.md` → `meta/handbook/`.

| Document | Owning leaf | Why |
|---|---|---|
| `PLAN-storage-locations.md` | `usage/storage/` | It is the implementation roadmap for the storage-location policy that `usage/storage/` already points at. |
| `PLAN-dbSendQueryArrow.md` | `usage/integrations/` | The DBI Arrow API is Arrow interchange, which is inside that leaf's scope line, and the stub already points at the plan. |
| `PLAN-vendoring-simplification.md` | `operations/vendoring/` | Its own scope is "the vendoring pipeline only", but it spans the model, the pipeline, the series loop, and the docs tree — no single child leaf covers it. Mapped to the area node, exactly as `scripts/VENDORING.md` already is. |
| `history/vendoring-loop.md` | `operations/vendoring/series-loop/` | It is the agentic-loop design that the series loop superseded, and it names the series-loop skill as its successor in its own first lines. |
| `history/main-dev-review-*.md` | `architecture/glue/` | Its findings are overwhelmingly C++ glue adaptations to upstream API changes; its own header disclaims being a description of the loop. Globbed rather than named so a future dated review of the same kind lands in the same group. |
| `README.md` | `meta/plans/` | The index itself, owned by the leaf that explains the directory. |

## Assumptions

- **`operations/vendoring/` is an internal node, not a leaf.** The generator only requires the owner path to exist, and `scripts/VENDORING.md` is already mapped to that node on `main`, so this follows precedent rather than inventing one. If the rule should be leaves-only, `PLAN-vendoring-simplification.md` would go to `operations/vendoring/pipeline/` and lose the series-loop and docs-tree halves of its scope.
- **`history/main-dev-review-2026-07.md` → `architecture/glue/` is the judgment call most worth a second opinion.** The alternative is `operations/vendoring/series-loop/` (it is a by-product of running the loop); I went with content over provenance.
- **Explicit filenames rather than a `PLAN-*.md` catch-all.** A new plan therefore lands under `Unowned` and fails `--check` until someone maps it. That is deliberate — it is the orphan rule made mechanical — but it does mean adding a plan requires touching the generator.
- **The old README's "what is current lives elsewhere" list** (`BRANCHES.md`, `RELEASE.md`, `scripts/VENDORING.md`, `scripts/EACH.md`, `.claude/skills/`) was not copied into the leaf. The fact it carried — the owner of the topic today wins over a plan — is kept; the list of who owns what is the tree's job and would rot as absorption proceeds. Each of those files is claimed by another leaf.
- **The purpose column for plan documents is each document's H1**, which is coarser than the old hand-written "To read about" phrases. No fact was lost that mattered: the one detail worth keeping — that the appendix of `history/vendoring-loop.md` is still cited by the cost model — is now stated in the leaf.
- **No plan document was moved or retired**, and `plan/PLAN-storage-locations.md` was deliberately left in place, to stay out of the concurrent `usage/storage/` rewrite.
- `handbook/meta/handbook/README.md` is another leaf's page; it was edited only for the one sentence this change makes false.

Verification: `docs-readme.R --check` is clean for both directories, `scripts/README.md` is byte-identical to `main`, and the repo-wide link check reports no dangling and no upward links.

## Durability sweep

A pass over the leaf for facts a routine commit would invalidate found nothing to
remove: the design of this change is what makes it immune. The leaf holds no
inventory — no count of plans, no count of history documents, no list of either —
because it opens by handing that question to the generated index at
[`plan/README.md`](plan/README.md), which is correct on the day it is read.

Kept deliberately: "the two live in separate directories" is the distinction the
section argues for, not a tally — a plan may still come true, a history never will —
and changing that number would mean changing the design. The named examples
(`usage/storage/` at the storage-location plan, `usage/integrations/` at the DBI
Arrow API's) illustrate the two-way pointer rather than enumerate it. The reference
to the ccache measurements in the appendix of `plan/history/vendoring-loop.md` names
where the numbers live and does not reproduce them.

The generator was not touched in this pass; `docs-readme.R --check` was re-run and
is clean for both `scripts/` and `plan/`.

One count in this PR description was reframed for the same reason — the ownership
map's rationale for `history/main-dev-review-*.md` no longer quotes a
finding-by-finding tally that only holds for one dated review.

---


krlmlr:
`main` now carries a written `meta/plans/` leaf and a `plan/README.md` that names every document under `plan/` (#2493). This branch predates that, conflicts throughout, and merging it would revert the newer tree. Closing unmerged.

Nothing verified about the code was salvaged from here — it proposed an organisation for `plan/`, not a claim to check. Its one lasting idea, a `history/` directory for records of work that is over rather than proposals for work, is where the wave's salvage document now lives; the rest of its layout (a section per owning handbook leaf) is not what `plan/README.md` settled on.

---

---

PR #2460, opened 2026-08-02
docs(handbook): Write the storage locations leaf

krlmlr:
**What this leaf now owns**

`handbook/usage/storage/` owns the storage-location topic in full — it is no
longer a pointer. A reader can answer "where do my extensions live and how do I
change that" from the leaf alone: what is written where (extension binaries,
stored secrets, the home setting the package deliberately never touches, logs,
the database file), the complete resolution chain for the extensions/secrets
home (`shared_home`, `home`, `duckdb.home`, `DUCKDB_R_HOME`, an existing
`~/.duckdb`, the interactive one-time offer, the per-session `tempdir()`), how
`config` beats all of it, how to move the store and what moving does *not* carry
over, the connect-time message and its two throttles, and the CRAN Repository
Policy reasoning that produced the design. It ends with the limits and the
neighbouring topics owned elsewhere (`usage/extensions/`, `usage/memory/`,
`usage/connections/`).

`?duckdb_storage` is unchanged and undiminished. It stays the user-facing
surface that ships in the tarball, and the leaf says in its second paragraph
that the reference page is derived from the leaf rather than the reverse —
which is what the rules already say about reference pages.

**What moved**

* `R/storage.R` — gains a plain `# Handbook: handbook/usage/storage/` comment
  immediately above the `duckdb_storage` roxygen block. Not a `#'` line:
  `handbook/` is `.Rbuildignore`d, so an `.Rd` link there would be broken for
  users. The roxygen itself is untouched, and `man/duckdb_storage.Rd` was not
  regenerated (a plain comment does not change it).
* `R/storage-status.R` — gains the same comment. Its roxygen feeds the same
  reference page via `@rdname`, so it is part of that page's source.
* `plan/PLAN-storage-locations.md` — gains the visible italic handbook
  backreference line under its H1. Nothing else in `plan/` is touched (#2459 is
  restructuring that directory).

**Verification**

Every claim is checked against `R/storage.R`, `R/storage-home.R`,
`R/storage-status.R`, `R/Driver.R`, `R/extensions.R`, the tests under
`tests/testthat/test-storage-*.R`, and the engine's
`src/duckdb/src/main/extension/extension_install.cpp` — which is where the
&lt;duckdb-version&gt;/&lt;platform&gt; path components under
`extension_directory` and the create-on-install behaviour come from. Nothing
was taken from the plan on trust. No build was run.

The plan is ahead of the code in four ways, and the leaf states each as a limit
rather than a fact:

* `duckdb_extension_storage()` / `duckdb_secret_storage()` and the
  `"session"`/`"user"`/`"shared"`/`"library"` vocabulary are ticked as done in
  the plan but exist in neither `R/` nor `NAMESPACE`. `home` and `shared_home`
  are the whole interface.
* The `.duckdb-r-keep` marker scheme and the package-library writability probe
  did not land — no marker string appears anywhere in `R/`.
* Migration between locations does not exist; changing the location leaves the
  old contents behind.
* The reactive ephemeral warning is deferred (the plan says so, and no such code
  exists).

**Assumptions**

* I widened the scope sentence to take logs and profiling output, which
  `?duckdb_storage` covers but no leaf in the tree claimed. They get one
  paragraph: the package sets none of `log_query_path`, `http_logging_output`,
  or the profiling output path (verified by grep over `R/` and the non-vendored
  `src/`), so nothing is written unless the user configures a path. If you would
  rather they went to a future logging leaf, this is the paragraph to move.
* I did **not** take `temp_directory` and the spill directory, even though
  `resolve_temp_directory()` lives in `R/storage-home.R` alongside the rest —
  `usage/memory/`'s scope line explicitly claims `temp_directory` and spill.
  The leaf names the setting and its two overrides only to point there. Two
  leaves owning it would violate "leaves explain, once"; if you prefer it here,
  `usage/memory/`'s scope line needs narrowing in the same change.
* The leaf says the package creates only the shared home (`~/.duckdb`, via
  `shared_home = TRUE` or an accepted prompt) and that the engine creates
  everything else on first write. `?duckdb`'s `@param home` says a `home` path
  is created "if needed", which is true in effect but not by R code — the only
  `dir.create()` calls in `R/` are the `~/.duckdb` ones.
* `plan/PLAN-storage-locations.md` is now largely realised, and its stale ticks
  are actively misleading. My reading is that it should retire to
  `plan/history/` once #2459 lands, but per instructions I left it in place and
  only added the backreference. Worth a maintainer decision.
* `?duckdb_storage` and the leaf now say the same things in two voices. That is
  the intended shape (leaf = truth, reference page = shipped surface), but it is
  a duplication that will need keeping in step; the backreference comment in
  `R/storage.R` is what makes the coupling discoverable.

**Durability sweep**

*Out, replaced by the mechanism:* the connect-time message's throttle figures.
"repeats at most once every eight hours" and "shown at most sixty times in a
session" became `STORAGE_MESSAGE_INTERVAL` and `STORAGE_MESSAGE_MAX` in
`R/storage-home.R`, which is where a reader gets the current values. The
durable fact — interactive throttling is by elapsed time, non-interactive by
count, because the two audiences can do different things about the message —
is stated more plainly than before.

*Reworded so it does not hinge on a count:* "The package creates a directory in
exactly one situation" is now "The package creates only the shared home
itself … The engine creates the rest of the tree on first write." The claim
was always about *who* creates what; a second `dir.create()` should not be able
to falsify it on a technicality.

*Cut as redundant:* "each of the two is decided independently" (the two
settings are named in the same sentence) and "Three neighbouring topics are
owned elsewhere" (the three are then named and linked).

*Kept:* the numbered resolution chain — those numbers are precedence, i.e. the
design, not a count. `dbdir`, `n = -1`-style code values, and the setting names
in the table are content. No links or structure changed.

---

### DIFF
--- a/R/storage.R
+++ b/R/storage.R
@@ -23,6 +23,7 @@
 # The package's tests and runnable examples also avoid the bundled C++ engine on
 # CRAN; see the CRAN guard in tests/testthat.R.
 
+# Handbook: handbook/usage/storage/
 #' DuckDB file-system usage: storage locations and how they are resolved
 #'
 #' @description
--- a/handbook/usage/storage/README.md
+++ b/handbook/usage/storage/README.md
@@ -95,7 +95,7 @@ DuckDB ecosystem.
 An `extension_directory` or `secret_directory` passed directly in the `config`
 list wins over all of this:
 the package fills in only the settings the caller did not supply,
-and each of the two is decided independently.
+and each is decided independently of the other.
 
 Because the decision is remade for each new instance,
 creating `~/.duckdb` — or setting the option or the variable — takes effect
@@ -135,11 +135,10 @@ to a newly chosen location:
 extensions are simply re-downloaded on demand,
 and secrets have to be recreated.
 
-The package creates a directory in exactly one situation —
+The package creates only the shared home itself —
 `~/.duckdb`, when `shared_home = TRUE` or when the interactive prompt is
 accepted.
-Every other directory in the tree is created by the engine
-when it first writes there.
+The engine creates the rest of the tree on first write.
 
 ## The connect-time message
 
@@ -152,11 +151,14 @@ where nobody was asked.
 
 Throttling differs by session type,
 because the two audiences can do different things about it.
-Interactively the message repeats at most once every eight hours;
+Interactively it is throttled by elapsed time —
+`STORAGE_MESSAGE_INTERVAL` in
+[`R/storage-home.R`](/R/storage-home.R) sets the gap;
 a human can act on it, and a gentle cadence fits.
-Non-interactively it is shown at most sixty times in a session and then
-goes silent for good, the last one saying so —
-a long-running or automated process should not be reminded forever.
+Non-interactively it is throttled by count instead —
+`STORAGE_MESSAGE_MAX` messages in a session, and then silence for good,
+the last one saying so —
+because a long-running or automated process should not be reminded forever.
 
 The message is suppressed whenever the location was chosen rather than
 inferred: by `home` or `shared_home`, by the `duckdb.home` option,
@@ -219,7 +221,7 @@ Specifically, the following are *not* available today:
   was deferred as too complex to land;
   the connect-time message above is what exists instead.
 
-Three neighbouring topics are owned elsewhere.
+Neighbouring topics are owned elsewhere.
 Which extensions ship and where an extension is installed *from* belongs to
 [`/handbook/usage/extensions/`](/handbook/usage/extensions/).
 The spill directory — `temp_directory`, its override via the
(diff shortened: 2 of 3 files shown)


krlmlr:
`main` now carries a written leaf for storage locations (#2493), which hands the topic to `?duckdb_storage` — the reference page ships in the tarball and owns the resolution order — and keeps only the two facts a reader must not miss. This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Salvaged from here, as a defect rather than as prose: `plan/done/PLAN-storage-locations.md` still marks the Phase 2 `.duckdb-r-keep` marker scheme and the library-writability probe `[x]`, and neither shipped. The plan's header already corrects two other unshipped items; this is a third.

Not salvaged: the resolution order, the message throttling, and the relocation-API absences, all of which `?duckdb_storage`, `R/Driver.R`'s docs, and the plan's own header already state.

---

---

PR #2463, opened 2026-08-02
docs(handbook): Write the series-loop leaf

krlmlr:
## What this leaf now owns

`handbook/operations/vendoring/series-loop/` explains the routine that vendors every series:
what fires it, what each of its stages decides, which playbook carries each move, and where the loop's limits are.
It stays a **pointer leaf** — the procedures remain machine-loaded under `.claude/skills/` and are linked, never copied.
The page states plainly that no workflow drives the loop (there is no vendoring workflow in the repository at all) and separates the repository's adjacent schedules — `sync.yaml`, `rcc-logs.yaml`, and `each.yaml` on pushes to `*-dev` — from the firing itself.
Boundaries are linked rather than absorbed: red runs to `vendoring/troubleshooting/`, the vendoring scripts to `vendoring/pipeline/`, shard mechanics and the verdict store to `ci/per-commit/`, the series refs to `branches/model/` and `branches/invariants/`, PR flow to `operations/review/`, and intent to `plan/PLAN-vendoring-simplification.md`.

## What moved

* `.claude/skills/series-loop.md` — gained a visible italic handbook backreference under its H1. Nothing removed; it stays the machine-loaded procedure.
* `.claude/skills/series-forward.md` — same backreference line.
* `.claude/skills/series-rebase.md` — same backreference line.
* `.claude/skills/series-open.md` — same backreference line.
* `AGENTS.md` — lost a stale schedule claim (vendoring does not run "daily via GitHub Actions"; the bullet now routes to the leaf that owns when the loop fires), and its routing row for the playbooks now names the leaf instead of the four skill files, which the leaf links in turn.
* `scripts/README.md` — unchanged. The generated index already groups `series-*.sh` under this leaf, so those scripts need no backreference of their own and the file-to-leaf mapping needed no edit.

Every behavioral claim on the page was checked against the source: `series-advance.sh` consumes at most one chunk per firing (`chunk=${2:-100}`), `series-check.sh` prints ADVANCE / WAIT / RETRY / REPAIR / IDLE plus a CUTOVER suggestion, `series-port.sh` classifies TOOLING / MIXED / OTHER / VENDOR and applies all but VENDOR, `series-cutover.sh` fails before the fetch when there is no terminal and takes the series name as confirmation.

## Assumptions

* **All four skills belong to this leaf.** The stub's `Today:` section listed `series-loop.md` plus the three siblings, and no other stub in the tree mentions `.claude/skills/`, so all four got the backreference. `series-open.md` is the closest call — opening a series is arguably `branches/model/` territory — but the loop is what discovers and then serves a new series, so it is listed here as the once-per-series playbook.
* **The schedule is genuinely outside the repository.** Nothing under `.github/workflows/` fires the loop and there is no `vendor.yaml`; `BRANCHES.md` and `plan/PLAN-vendoring-simplification.md` both call it "a scheduled Claude routine". The page says the schedule lives with the routine that invokes the playbook, outside this tree, rather than inventing a cadence. Confirm that is the intended framing.
* **`AGENTS.md`'s vendoring bullet was in scope.** Most of that section belongs to the `model/` and `pipeline/` leaves being written in parallel, so only the one bullet carrying the (false) trigger claim was touched. If the pipeline leaf rewrites the whole section, this edit is a trivial merge.
* **`BRANCHES.md` was left alone, and the contradiction it carries is now visible in the tree.** Its "Synchronization › Vendoring" paragraph says `-dev` is consumed "up to 25 commits at a time", which contradicts the verified default in `series-advance.sh` (`chunk=${2:-100}`). Since the leaf now states the hundred in prose (see the sweep below), the two documents disagree in the open. That file is owned by the `branches/` leaves, so the discrepancy is reported rather than fixed here.
* **Ref mechanics are named, not explained.** The series refs are named once so the stages read, with the model itself left to `branches/model/`; a reviewer should check the split lands where they expect.

## Durability sweep

**Out, replaced by the mechanism:**

* "`sync.yaml` … hourly" and "`rcc-logs.yaml` … twice an hour" → "each on the
  schedule its own `cron:` line sets". Both workflows are already linked from
  that sentence, so a reader gets today's cadence from the file. The argument
  the sentence makes — those cadences bound how fast a firing can learn
  anything, but do not decide when it fires — is untouched.

**Out, because the count served no argument:** "what each of the four
guarantees" → "what each of them guarantees". The sentence only hands the refs
to `branches/model/`; counting them there and here invites the two to disagree.

**Kept:** the stage numbers 0–7. They are identifiers, not a tally — the leaf
says "stages 3 and 5" and "stage 4", and the playbooks use the same numbering.
The literal classifications `TOOLING` / `MIXED` / `OTHER` / `VENDOR` are values,
not a count. The single-retry rule ("the one rerun was spent") is the policy
itself.

## Correction after the sweep

The sweep had also replaced the extend cap, "up to 100 commits per firing",
with "one chunk per firing, sized by `series-advance.sh`'s chunk argument".
That went too far, and a follow-up commit put the value back. A default
governs behaviour, so a reader who must open the script to learn it has been
sent away for the thing they came for.

Stage 5 now reads **"one chunk per firing — a hundred commits by default,
overridable per call"**, and the sentence after it delegates the bound's
rationale — what the bound buys — to
[`branches/model/`](/handbook/branches/model/) rather than arguing it here.
The mechanism stays beside the value: the default is `chunk=${2:-100}` in
`series-advance.sh`. Only the chunk size was restored; the two cadence
pointers above stand as the sweep left them.

---

### DIFF
--- a/.claude/skills/series-forward.md
+++ b/.claude/skills/series-forward.md
@@ -1,5 +1,7 @@
 # Forwarding a series to a newer base
 
+*Handbook: [`operations/vendoring/series-loop/`](/handbook/operations/vendoring/series-loop/README.md).*
+
 `main` moves under a series:
 R-side fixes, workflow changes, version bumps.
 The series' branches are built on yesterday's `main`,
--- a/.claude/skills/series-loop.md
+++ b/.claude/skills/series-loop.md
@@ -1,5 +1,7 @@
 # The series loop: vendor, promote, repair
 
+*Handbook: [`operations/vendoring/series-loop/`](/handbook/operations/vendoring/series-loop/README.md).*
+
 One routine drives **all** series:
 each firing enumerates every series and every existing forward (`-fwd`)
 counterpart and serves each in turn.
--- a/.claude/skills/series-open.md
+++ b/.claude/skills/series-open.md
@@ -1,5 +1,7 @@
 # Opening a new series
 
+*Handbook: [`operations/vendoring/series-loop/`](/handbook/operations/vendoring/series-loop/README.md).*
+
 When upstream cuts a release branch —
 v2.0 is released, upstream `main` becomes the 2.1 line —
 the release gets a series of its own.
--- a/.claude/skills/series-rebase.md
+++ b/.claude/skills/series-rebase.md
@@ -1,5 +1,7 @@
 # Rebasing a forward series onto a newer mainline
 
+*Handbook: [`operations/vendoring/series-loop/`](/handbook/operations/vendoring/series-loop/README.md).*
+
 Two different moves put a series on a newer base,
 and only one of them is a rebase.
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,7 +16,7 @@ start there rather than searching.
 | Release process, modelled as a state machine | `RELEASE.md` |
 | Vendoring mechanics: scripts, invariants, troubleshooting | `scripts/VENDORING.md` |
 | Per-commit CI (sharded matrix): design and limits | `scripts/EACH.md` |
-| Operating the vendoring loop (routine playbooks) | `.claude/skills/`: `series-loop.md`, `series-forward.md`, `series-rebase.md`, `series-open.md` |
+| Operating the vendoring loop: what fires it, what its stages decide | [`handbook/operations/vendoring/series-loop/`](/handbook/operations/vendoring/series-loop/), which links the playbooks in `.claude/skills/` |
 | Designs, plans, and historical documents | `plan/README.md`, which names each by path |
 
 ## Working Effectively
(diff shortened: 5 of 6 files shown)


krlmlr:
`main` now carries a written `operations/vendoring/series-loop/` leaf (#2493), and the loop's stages are owned by the machine-loaded playbooks under `.claude/skills/`. This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Nothing was salvaged from here. What it added was a retelling of the skills' own stages and of the refs `branches/model/` owns — the playbook is the artifact, and the leaf's job is to say which one runs when.

---

---

PR #2464, opened 2026-08-02
docs(handbook): Write the testing/snapshots leaf

krlmlr:
## What this leaf now owns

`handbook/testing/snapshots/` explains snapshot testing in this package:
what `tests/testthat/_snaps/*.md` records and why so much of it is engine output rather than R output;
the package-name normalisation in `tests/testthat/helper-snapshot.R` that keeps one recorded file valid on every flavor, given that `scripts/lts.patch` deliberately leaves `_snaps/` alone;
the three routes to accepting a changed snapshot (by hand, the `update-snapshots` composite action in CI, and `scripts/snapshot-accept.sh` for one commit deep in a series);
and the judgment no script header stated — when accepting is right, and the six shapes of diff where accepting hides a regression.
Running the suite stays [`testing/suite/`](/handbook/testing/suite/README.md)'s topic, the CRAN gate stays [`testing/guards/`](/handbook/testing/guards/README.md)'s, and consuming snapshot branches during vendoring stays [`operations/vendoring/series-loop/`](/handbook/operations/vendoring/series-loop/README.md)'s; all are linked, not repeated.

## What moved

- `scripts/snapshot-accept.sh` — lost the two explanatory paragraphs of its header: its role as the fallback for snapshot repair when no snapshot-&lt;sha&gt;-rcc-smoke-null branch exists or its diff does not survive review, and the note that it leaves the corrected files in the working tree for the caller to fold. Both now live in the leaf. It keeps the extractable one-line summary that `scripts/README.md` is built from (#2453), plus its usage block. It gained nothing: `scripts/` has a generated index that already carries the backreference.

`scripts/README.md` is unchanged and verified current (`Rscript .claude/skills/docs-consistency/docs-readme.R --check`) — the `globs =` mapping in `.claude/skills/docs-consistency/docs-readme.R` already assigns `snapshot-accept.sh` to `testing/snapshots/`, and shrinking the header did not touch its first sentence.

## Assumptions

- The stub's "absorb the procedure in the header" was read as *move the explanation, keep the operating instructions*: the one-line summary and the `Usage:` block stayed in the script, everything else moved. If the intent was to strip the usage too, say so.
- The `update-snapshots` action's own comment says a pull request "will use reviewdog/action-suggester in the commit action". That is stale — `.github/workflows/commit/action.yml` contains no reviewdog step; reviewdog only appears in `format-suggest.yaml`. The leaf therefore states what the code actually does on a PR: the accepted files stay in the working tree, and the later `commit` step pushes them onto the branch, or uploads a patch artifact and fails for a fork PR. Worth confirming that the stale comment (not the leaf) is what should be fixed.
- The action is invoked from `R-CMD-check.yaml` (both `rcc-smoke` and `rcc-full`) and `R-CMD-check-dev.yaml`, with slightly different fork guards per job. The leaf says "the R-CMD-check jobs" and links [`operations/ci/workflows/`](/handbook/operations/ci/workflows/README.md) rather than enumerating them, on the grounds that the workflow inventory is that leaf's topic.
- "The output is not deterministic" is grounded in a real case from `plan/history/main-dev-review-2026-07.md`: a one-shot nanosecond-coercion warning (`std::call_once`) made the `types.md` snapshot order-dependent, and the fix was to exclude the type from `test_all_types()`, not to re-record. The leaf states the hazard generically without linking that historical review, since it is a review record rather than intent.
- No `plan/` document carries intent for snapshots, so the leaf points at none.
- ~~The suite has 63 `expect_snapshot()` calls and one `expect_snapshot_error()` across 12 recorded files; the leaf says "about sixty" so the number cannot rot.~~ Superseded by the durability sweep below — the count is gone from the leaf entirely.

## Durability sweep

A targeted pass for one defect class: facts a routine commit invalidates.

**Removed.** "About sixty `expect_snapshot()` calls (and one
`expect_snapshot_error()`)" is now plainly "`expect_snapshot()` calls".
Rounding to "about sixty" did not save it: the next contributor who adds a
snapshot file makes it wrong anyway, and the figure carried no argument. The
sentence exists to say where the expectations live and where the text they
capture is recorded, and it still names both `tests/testthat/` and the
per-test-file Markdown under `tests/testthat/_snaps/` — so a reader who wants
the number counts it, and is right on the day they look.

**Kept.** "One Markdown file per test file, one section per test" is the
recording format, not a tally; it is what a reader needs in order to find a
given snapshot. The six shapes of hiding-a-regression diff are a taxonomy of
judgment, not a count of things in the tree.

---

### DIFF
--- a/handbook/testing/snapshots/README.md
+++ b/handbook/testing/snapshots/README.md
@@ -1,17 +1,148 @@
 # Snapshots
 
-*Stub — this leaf will own its topic;
-today it routes to where the knowledge lives.
-The writing protocol is in [`meta/handbook/`](../../meta/handbook/);
-the last section holds this leaf's parameters.*
+How the suite records expected output as snapshot files,
+and how a changed snapshot is accepted —
+deliberately, because accepting one asserts that the new output is correct.
 
-Scope: snapshot tests, and accepting changed snapshots deliberately.
+## What a snapshot records
 
-Today:
+Snapshot expectations are testthat's.
+About sixty `expect_snapshot()` calls (and one `expect_snapshot_error()`)
+are spread over `tests/testthat/`,
+and the text they captured lives in `tests/testthat/_snaps/<file>.md` —
+one Markdown file per test file,
+one section per test, holding the code and its printed output.
+Running them is running the suite
+([`testing/suite/`](/handbook/testing/suite/README.md));
+whether they run at all under CRAN's rules is
+[`testing/guards/`](/handbook/testing/guards/README.md).
 
-* [`scripts/snapshot-accept.sh`](../../../scripts/snapshot-accept.sh) — its header is the procedure
+Much of the recorded text comes from the engine rather than from R:
+error messages and the position they point at, type names,
+`EXPLAIN` plans, the storage summary.
+A snapshot therefore drifts whenever a vendored DuckDB changes its output,
+which is why snapshot repair is a routine part of vendoring
+([`operations/vendoring/series-loop/`](/handbook/operations/vendoring/series-loop/README.md)).
 
-To write this leaf:
+The one thing a snapshot must not record is the package's own name.
+Every flavor but the mainline one ships under a different name
+([`branches/flavors/`](/handbook/branches/flavors/README.md)),
+and anything derived from `get_package_name()` carries that name into
+the output it produces —
+the session storage home most visibly.
+`scripts/lts.patch` deliberately does not rewrite `tests/testthat/_snaps/`,
+so a file recorded on one flavor could never match another.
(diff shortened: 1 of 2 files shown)


krlmlr:
`main` now carries a written `testing/snapshots/` leaf (#2493), with the flavor rule, the three acceptance routes, and the test that a diff must read as the change that caused it. This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Nothing was salvaged from here. Its substantial addition was a checklist of ways accepting a snapshot can hide a regression; those are review judgement rather than facts about the code, and the one that was a code claim — a non-deterministic one-shot warning whose snapshot depends on test order — named no file that proves it, so it did not survive verification.

---

---

Issue #2365, opened 2026-06-20
Investigate jemalloc


krlmlr:
Entered our build system with https://github.com/duckdb/duckdb/pull/22811 (in v2.0.0) and broke the build in https://github.com/krlmlr/duckdb-r/commit/34f2add5658e74fb070fe7675964dbdbedbaded. Need to understand the tradeoff.

krlmlr:
@staticlibs: I investigated a bit.

Can you confirm that jemalloc is Linux + glibc only? Enabling it for Linux only is a bit more difficult than if it were on for all platforms, still doable. We probably don't want `OVERRIDE_NEW_DELETE`, and test very carefully.

How to measure the effect? What benchmark results would improve if we enabled jemalloc?

staticlibs:
@krlmlr 

Hi Kirill!

> Can you confirm that jemalloc is Linux + glibc only

Linux (both glibc and musl) and macOS. We also intend to enable it on Windows, but right now it is not enabled there.

> We probably don't want OVERRIDE_NEW_DELETE

Yes, it may be better to exclude `jemalloc_cpp.cpp` ([JDBC example](https://github.com/duckdb/duckdb-java/blob/a487003012d64c04a583eb6188e69068006a6591/vendor.py#L49)).

> How to measure the effect? What benchmark results would improve if we enabled jemalloc?

Sorry, cannot help with this bit.

Issue #1829, opened 2025-12-17
g++ with std=gnu++17 warns about declared but unused variables in the underlying DuckDB library


pachadotdev:
Here is the log

```
 In file included from duckdb/src/include/duckdb/common/sort/duckdb_pdqsort.hpp:29,
                    from duckdb/src/common/sort/radix_sort.cpp:3,
                    from duckdb/ub_src_common_sort.cpp:7:
   duckdb/src/include/duckdb/common/operator/numeric_cast.hpp:584:6: warning: ‘bool duckdb::TryCastWithOverflowCheck(SRC, DST&) [with SRC = double; DST = uhugeint_t]’ defined but not used [-Wunused-function]
     584 | bool TryCastWithOverflowCheck(double value, uhugeint_t &result) {
         |      ^~~~~~~~~~~~~~~~~~~~~~~~
   duckdb/src/include/duckdb/common/operator/numeric_cast.hpp:579:6: warning: ‘bool duckdb::TryCastWithOverflowCheck(SRC, DST&) [with SRC = float; DST = uhugeint_t]’ defined but not used [-Wunused-function]
     579 | bool TryCastWithOverflowCheck(float value, uhugeint_t &result) {
         |      ^~~~~~~~~~~~~~~~~~~~~~~~
   duckdb/src/include/duckdb/common/operator/numeric_cast.hpp:574:6: warning: ‘bool duckdb::TryCastWithOverflowCheck(SRC, DST&) [with SRC = long unsigned int; DST = uhugeint_t]’ defined but not used [-Wunused-function]
     574 | bool TryCastWithOverflowCheck(uint64_t value, uhugeint_t &result) {
         |      ^~~~~~~~~~~~~~~~~~~~~~~~
   duckdb/src/include/duckdb/common/operator/numeric_cast.hpp:569:6: warning: ‘bool duckdb::TryCastWithOverflowCheck(SRC, DST&) [with SRC = unsigned int; DST = uhugeint_t]’ defined but not used [-Wunused-function]
     569 | bool TryCastWithOverflowCheck(uint32_t value, uhugeint_t &result) {
         |      ^~~~~~~~~~~~~~~~~~~~~~~~
   duckdb/src/include/duckdb/common/operator/numeric_cast.hpp:564:6: warning: ‘bool duckdb::TryCastWithOverflowCheck(SRC, DST&) [with SRC = short unsigned int; DST = uhugeint_t]’ defined but not used [-Wunused-function]
[...]

PR #2466, opened 2026-08-02
docs(handbook): Write the engine leaf

krlmlr:
## What this leaf now owns

`handbook/architecture/engine/` now explains what is *ours* about the DuckDB engine embedded in `src/duckdb/`, and points outward for everything else. It covers which commit is embedded (`DUCKDB_VERSION` and `DUCKDB_SOURCE_ID` in the vendored `pragma_version.cpp`, both visible through `PRAGMA version`), how that version reaches R (`rconfigure.py` generates `R/version.R`, which `dbGetInfo()` reports without opening a database), and how this build differs from a stock build: a table of the `-D` flags in `src/Makevars.in` with the effect of each, the vendor-time environment variables, the linked extension set (`parquet` and `core_functions` only, fixed at vendor time), the jemalloc exclusion, and the warnings the vendored tree raises. Engine internals stay at duckdb.org and `duckdb/duckdb`; maintenance of the copy stays at `operations/vendoring/`; the shipped extension set as a user-facing fact stays at `usage/extensions/`; the no-suppression policy stays at `architecture/glue/`.

Both listed issues are drained in prose and stay open, since each would close on a code change rather than on words: [#2365](https://github.com/duckdb/duckdb-r/issues/2365) (jemalloc is excluded today and `DUCKDB_ENABLE_JEMALLOC` is never defined; enabling it deliberately waits on a benchmark) and [#1829](https://github.com/duckdb/duckdb-r/issues/1829) (warnings from the vendored tree are emitted, not silenced, and the shipped build sets no `-Werror`).

## What moved

Nothing was absorbed from a free-floating document, so nothing was deleted. The changes outside `handbook/` are backreferences:

- `R/version.R` — gained a `# Handbook: handbook/architecture/engine/` comment.
- `scripts/rconfigure.py` — its template for `R/version.R` gained the same line, so the next vendor run reproduces the comment instead of dropping it. Nothing else about the script changed, and no backreference was added for the script itself; it belongs to `operations/vendoring/pipeline/`.
- `src/Makevars.in` — gained a comment naming both leaves that serve it: this one for the `-D` flags, `handbook/build/source-build/` for the file itself.
- `src/Makevars`, `src/Makevars.win` — the same line, at the position a regeneration from `Makevars.in` would place it, so the generated files do not drift from their template.

## Assumptions

- **`src/Makevars.in` is served by two leaves.** `build/source-build/` owns the file; this leaf owns the meaning of the `-D` flags in it. I wrote one backreference line naming both rather than leaving the flags without a route from the file. If the source-build leaf lands its own line, the merge should collapse them into one.
- **Hand-editing the generated `src/Makevars` and `src/Makevars.win`** is normally forbidden, but the added line comes from `Makevars.in` and is exactly what the next `rconfigure.py` run emits — so this keeps them in sync rather than breaking the invariant. Worth a look if a guard compares them.
- **Facts stated are read from the sources, not from a build** — `scripts/rconfigure.py`, `src/Makevars`/`.in`/`.win`, `configure`, `configure.win`, `R/version.R`, and the vendored headers that consume each flag. Nothing was verified by compiling, per the documentation-only scope of this wave.
- **The engine&#39;s own flavors of these facts were left out on purpose**: what a release `libduckdb` links instead is stated only as the reason engine configuration is never verified on the fast path, and is left to `build/fast-paths/` and `usage/extensions/`.
- **`-DDUCKDB_R_BUILD`** is described by its only effect in the vendored tree today (a non-atomic `next` pointer in `SegmentNode`). I did not find, and did not invent, a rationale for that substitution.

## Durability sweep

**Out:**

- *"links exactly two extensions, `parquet` and `core_functions`"* → the two names, plus the mechanism: the `-DDUCKDB_EXTENSION_*_LINKED` defines in `src/Makevars` are the authoritative list. A reader who needs today's set reads it there.
- *"some 1800 C++ source files"* → *"well over a thousand"*. The point is the scale that makes the warnings noise, and that survives an upstream bump.
- *"Neither of the first two"* → the flags it means, `-Werror` and `-DDUCKDB_DEBUG_MOVE`, so the sentence no longer depends on the order of the list above it.

**Kept:**

- The **two version identifiers**, `DUCKDB_VERSION` and `DUCKDB_SOURCE_ID` — changing the number would change the design, and both are named.
- **`v1.5.5`** — a version identifying the format of the release string, not a count.
- **"broke the build twice over"** — both breakages are named immediately after; a list that names its members, not a bare count.
- The `-D` flag table, which is an inventory rather than a count.

---


krlmlr:
`main` now carries a written leaf for the engine at reference depth (#2493), with the identity pair, the load-bearing flags, and the jemalloc exclusion. This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Salvaged from here: what `-DDUCKDB_R_BUILD` actually does. Its only effect on the vendored tree is to replace the `atomic` next pointer in `SegmentNode` with a plain one, in `segment_tree.hpp` — which makes it the flag whose name promises most and delivers least, and worth stating once rather than leaving a reader to grep for it.

Not salvaged: the jemalloc reasoning. `scripts/rconfigure.py`'s own header already carries it in full, which is where the ladder says it belongs.

---

---

PR #2468, opened 2026-08-02
docs(handbook): Write the build/fast-paths leaf

krlmlr:
## What this leaf now owns

`handbook/build/fast-paths/` explains linking a prebuilt `libduckdb`
instead of compiling the vendored sources:
the `DUCKDB_R_USE_SYSTEM_LIB` opt-in and how `scripts/install-libduckdb.sh`
resolves version, platform, and prefix;
the `pkgload::load_all()` / `testthat::test_local()` loop that honors the
same switch;
what the mode actually swaps (implementation only — the glue keeps
compiling against the vendored internal headers) and the
`src/Makevars.system-lib` + rpath mechanics;
the `configure` commit-match guard, including the ten-character hash
comparison and what it prints on a mismatch;
the CI default and the pinned `DUCKDB_R_USE_SYSTEM_LIB=0` matrix
entries that keep a vendored build on every OS;
and the limits — not for `R CMD build` or for users, no prebuilt between
releases, Linux and macOS only.
It states as a boundary that engine configuration is never verified on
the fast path, since the release `libduckdb` links extra extensions
(`icu`, `json`, `autocomplete`) and carries a different
`autoinstall_known_extensions` default, and links
`/handbook/usage/extensions/` for what the package actually ships.
[#22](https://github.com/duckdb/duckdb-r/issues/22) is answered in the
"Limits" section and left open for its user-facing half.

## What moved

* `AGENTS.md` — loses §§ "Fast build with system libduckdb (Linux/macOS,
  opt-in)" and "Testing with prebuilt DuckDB (the fast iterate loop)",
  each replaced by a one-line pointer to the leaf; gains the visible
  italic handbook backreference under its H1. Its cross-reference from
  "Bootstrap, Build, and Test the Repository" now links the leaf instead
  of the two removed anchors. The `.dd` dependency-file paragraph that
  sat inside the absorbed region stays put — it belongs to
  `build/source-build/`.

`scripts/README.md` needed no change: its generator already maps
`install-*.sh` to `handbook/build/fast-paths`, so the index carries that
backreference.

## Assumptions

* `AGENTS.md` claimed that development snapshots are fetched "from the
  DuckDB nightly staging bucket keyed by `DUCKDB_SOURCE_ID`". The code in
  `scripts/install-libduckdb.sh` does not do this: for a `*-dev*` version
  it prints that no prebuilt is published and exits 0 so the caller falls
  back to a source build. The leaf states the code's behavior. The
  script's own header comment (lines 13–17) still carries the stale
  claim; it is left untouched here, since this wave is documentation
  only and `scripts/` files take no hand-added backreference.
* `pkgload::load_all()` honoring the opt-in is stated on the mechanism —
  it compiles through the same `configure` and `src/Makevars`, and
  `src/Makevars` is what includes `Makevars.system-lib`. Not re-measured.
* `scripts/install-duckdb-cli.sh` also maps to this leaf through the
  generated index's `install-*.sh` glob, but is not a fast-path tool. The
  leaf names it in one sentence under "Limits" so a reader arriving from
  `scripts/README.md` is not stranded, rather than absorbing it; if the
  mapping should instead move to a testing leaf, that is a separate change.
* The root `README.md` was left alone: `AGENTS.md` is only partly
  absorbed and is not deleted, so its `## Documentation` entry still
  describes it correctly.
* No build was run, per the wave's rules; every behavioral claim is read
  off `configure`, `src/Makevars.in`, `scripts/install-libduckdb.sh`,
  `scripts/rconfigure.py`, `.github/versions-matrix.R`, and
  `.github/workflows/custom/before-install/action.yml`.

## Durability sweep

A follow-up pass removed the numbers a routine commit invalidates —
mostly figures this leaf had carried over unverified from `AGENTS.md`
and `configure`.

**Out**, replaced by the mechanism or by a qualitative form:

* "about 1700 vendored `.cpp` files", "10–15 minutes", "roughly five
  seconds" → the whole vendored tree, many minutes, seconds. The
  argument was always the order of magnitude, not the stopwatch.
* "~30 glue files" → "the glue translation units in `src/`".
* The `load_all()` measurement ("about a second", "about four") →
  "seconds rather than the minutes a vendored build takes".
* "about 71 internal DuckDB C++ headers", "roughly 37 of them" absent
  from `duckdb.hpp` → a pointer to the comment block at the top of
  `configure`, where the reasoning and the header wiring live. This also
  retires the assumption listed above: there is no longer a count to
  have recounted.
* "one Linux and one macOS entry" → "a Linux and a macOS entry"; the
  per-platform pinning is the fact, the entry count is incidental.

**Kept:** the ten/eleven hex characters of the commit-match comparison —
that is the guard's mechanism, and changing it would change the design.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,7 @@
 # DuckDB R Package - Operational Instructions
 
+*Handbook: [`build/fast-paths/`](/handbook/build/fast-paths/README.md).*
+
 R package that contains a vendored copy of the DuckDB C++ library and glue code for R, including a DBI and a relational interface.
 
 ## Where to look
@@ -35,7 +37,7 @@ so the check reports "Found no calls to `R_registerRoutines`, `R_useDynamicSymbo
 and fails a package that registers its routines correctly.
 It saves ~10-20 s on this tree, so it is not worth carrying into CI at all.
 
-For an interactive edit-build-test loop, prefer the prebuilt-libduckdb fast path (seconds instead of 10-15 minutes) -- see ["Fast build with system libduckdb"](#fast-build-with-system-libduckdb-linuxmacos-opt-in) and ["Testing with prebuilt DuckDB"](#testing-with-prebuilt-duckdb-the-fast-iterate-loop) below.
+For an interactive edit-build-test loop, prefer the prebuilt-libduckdb fast path (seconds instead of 10-15 minutes) -- see [`build/fast-paths/`](/handbook/build/fast-paths/README.md).
 
 ### Run Tests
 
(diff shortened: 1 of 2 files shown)


krlmlr:
`main` now carries a written `build/fast-paths/` leaf at core depth (#2493), with the opt-in, the commit-match guard, the `load_all()` equivalence, and the warning that the fast path proves nothing about the engine. This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Nothing was salvaged from here. Its limits section is the closest it came, and the newer leaf already states each of them: install-only and never for a tarball someone else installs, Linux and macOS only, and drop the variable when no prebuilt matches a `-dev` snapshot. The remaining item — that #22 stays open for the user-facing half of a prebuilt install — is the issue's to carry.

---

---

PR #2469, opened 2026-08-02
docs(handbook): Write the branches/flavors leaf

krlmlr:
## What this leaf now owns

`handbook/branches/flavors/` explains one source tree published under many
package names: why more than one package exists (CRAN's one-version-per-name
rule, the four-month upstream minor cadence, one-year LTS support, `.dev` for
pre-CRAN testing), the flavor set that exists today and where each is built
from, the exact file surface the rename touches (`DESCRIPTION`, the
`@useDynLib` / `@name` pair with its patched `NAMESPACE` and `.Rd`,
`DUCKDB_PACKAGE_NAME`, the renamed public `duckdb_*_types.hpp`, `testthat.R`,
the README blurb, and the regenerated cpp11 pair), and the tooling that
produces it — `scripts/flavor.sh` and `scripts/flavor.patch` — with its limits.
Boundaries are linked, not restated: the `get_package_name()` seam is
`architecture/r-layer/`'s, the guard that enforces it is `testing/guards/`'s,
which flavor a user should install is `usage/installation/`'s, the series and
their refs are `branches/model/`'s, and the name-coherence and mechanical-rename
invariants are `branches/invariants/`'s.

## What moved

- **`BRANCHES.md`** — lost §§ "Why multiple R packages?" and "R Package
  Flavors"; both headings remain and now hold a one-line pointer to the leaf, so
  the `#r-package-flavors` anchor that `AGENTS.md` links keeps resolving. The
  "On new minor release" step that said to update the flavors table in that
  document now points at the leaf. Gained the visible handbook backreference
  line under its H1.
- **`README.md`** — gained the visible handbook backreference line under its H1,
  and the `Flavors` badge legend now routes to the leaf for what a flavor is and
  how it is produced, keeping the `BRANCHES.md` link for the branch model. The
  table itself is untouched.
- **`scripts/README.md`** — unchanged. It is generated, and its `flavor*` glob
  group already maps to `handbook/branches/flavors`, so no ownership edit and no
  regeneration were needed.

## Assumptions

- "Replace each absorbed heading with a one-line pointer" was read as *keep the
  heading, replace its body*, rather than deleting the heading outright.
  Deleting it would break `AGENTS.md`'s `BRANCHES.md#r-package-flavors` link and
  the three sibling leaves splitting the same file.
- The root `README.md` gained a backreference line under its H1 even though
  `usage/installation/` also serves that file; per the wave convention the two
  leaves merge into one line. `git apply --check --include=README.md
  scripts/flavor.patch` and a full `patch -p1 --dry-run` still pass with the
  extra line, and `flavor_package_name_offenders(".")` stays empty.
- The flavor table's "Built from" column names the branch for the two published
  stable flavors (`main`, `v1.4-andium-lts`) but only names the *series* for the
  `.dev` flavors, deferring which ref r-universe builds to `branches/model/`.
  `BRANCHES.md` says `-dev` while `.claude/skills/series-loop.md` says `-green`
  is what r-universe should build from; rather than pick, the leaf links.
- Three facts are documented as today's behavior, verified from source rather
  than from `BRANCHES.md`: `flavor.sh` requires `gsed` (absent on a plain
  Linux); the substitution does not reach the `1.3` in the README blurb the
  patch inserts, so every flavored branch's README says "LTS version 1.3"
  (confirmed on the published `v1.4-andium-lts`); and the cpp11 0.5.5
  first-dot-only symbol hazard from `.claude/skills/series-open.md`. None of
  these is filed as an issue — flagging them here in case they should be.
- `BRANCHES.md`'s tooling table calls `1.4` the `flavor.patch` placeholder; the
  file and the script's own comment both say `1.3`. The leaf states `1.3`. That
  table belongs to another leaf, so it was left alone.
- No `plan/` document covers flavors, so the leaf points at no intent document.

## Durability sweep

**Out.** The opening no longer repeats every r-universe package name; it says
`duckdb` on CRAN plus numbered and `.dev` names on r-universe, and points at
the flavor table below. The set was written down twice, so a new series had to
be added in two places. The renamed public header is now shown with one
example (`duckdb.1.5.dev` → `duckdb_1_5_dev_types.hpp`) instead of one
filename per flavor, which was the same list in a third place.

**Stayed.** The flavor table itself — it names each flavor and the branch it is
built from, which is the *which*, not a count, and the leaf already says the
set grows when a series opens and where a new one is announced by hand. The
version identifiers `v1.4-andium-lts`, `cpp11 0.5.5` and the `1.3` placeholder
in `flavor.patch`, which identify a thing. "Two commits" and "the two cpp11
files" in `flavor.sh`'s description: the numbered list is the design, and
changing it would mean changing the script. The upstream cadence and LTS
support window, which are upstream policy rather than readings off this tree.

---

### DIFF
--- a/BRANCHES.md
+++ b/BRANCHES.md
@@ -1,5 +1,7 @@
 # Branching Strategy
 
+*Handbook: [`branches/flavors/`](/handbook/branches/flavors/README.md).*
+
 Version numbers are given at the time of writing (March 2026) and may be outdated by the time you read this.
 The branching strategy is expected to remain stable.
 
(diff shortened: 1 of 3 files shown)


krlmlr:
`main` now carries a written leaf for flavors at core depth (#2493), with the flavor table and the rename surface. This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Salvaged from here as unfixed defects rather than as prose: `scripts/flavor.sh` invokes `gsed`, so it fails on any system where GNU sed is simply `sed`; and the rename never reaches the README blurb — `flavor.patch` writes "This package contains the LTS version 1.3 of DuckDB", and `flavor.sh`'s substitution requires a `.` or `_` immediately before the `1`, so the space-separated `1.3` is never rewritten and every flavored branch announces version 1.3 whatever it carries. Visible today on the published `duckdb.1.4`.

Both are in the defects section of `plan/history/2026-08-handbook-wave-salvage.md` — #2507, [file](https://github.com/duckdb/duckdb-r/blob/claude/handbook-wave-salvage-zayxqn/plan/history/2026-08-handbook-wave-salvage.md).

---

---

PR #2472, opened 2026-08-02
docs(handbook): Write the vendoring troubleshooting leaf

krlmlr:
## What this leaf now owns

`handbook/operations/vendoring/troubleshooting/` is the page you open when a
vendoring run is red. It is organized by symptom rather than by subsystem:
where to look first (`series-check.sh`, the harvest on `rcc`, what is
vendored), a script refusing to start, an empty base scan, the glue gate
stopping the walk, a patch that disappeared from `patch/`, a red commit on
`{S}-dev` and how its log classifies, a commit with no verdict at all, diffs that
are not changes, re-vendoring from scratch, and the limits. Every claim is
checked against `scripts/vendor.sh`, `scripts/vendor-one.sh`,
`scripts/series-check.sh`, `scripts/rcc-one.sh`, `scripts/rcc-decided.sh`,
`src/include/deps.mk` and `.github/workflows/each.yaml`. Stuck shards and the
verdict store point at `operations/ci/per-commit/`, snapshot acceptance at
`testing/snapshots/`, the patch-stack lifecycle at `operations/vendoring/pipeline/`,
and the repair playbooks at `operations/vendoring/series-loop/` — linked, not
absorbed.

(Series refs are written `{S}-dev` here. The leaf itself uses angle brackets;
GitHub deletes those from a pull-request body even inside backticks.)

## What moved

- **`scripts/VENDORING.md`** — lost its whole `## Troubleshooting` section
  ("Vendoring stopped working", "Manual recovery", "Common issues") and the
  diagnostic half of `## Monitoring Vendoring` ("GitHub Actions",
  "Commit history", "Version tracking"); each is replaced by a one-line pointer
  to the leaf. `### Ahead/behind badges` stays untouched, and the intro sentence
  now routes troubleshooting to the handbook instead of claiming it. No
  backreference line was added: `VENDORING.md` is covered by the generated
  `scripts/README.md` index, whose ownership row is unchanged (verified with
  `docs-readme.R --check`).

`scripts/EACH.md` was **not** edited — another leaf owns it.

## Assumptions

- **`## Monitoring Vendoring` was split rather than taken whole.** The work
  order named it as a source, but `### Ahead/behind badges` is about series
  display and `README.md` upkeep (shields.io within one repository, release
  branches mirrored into the fork), not about a red run, so it was left in
  `VENDORING.md` for whichever leaf claims it. The three diagnostic subsections
  were absorbed.
- **The retired `vendor-gate.sh` verdicts were replaced, not repeated.**
  `VENDORING.md` still spoke of a gate saying `red` or `stale`; that script no
  longer exists in the tree. The leaf states today's verdicts —
  `series-check.sh`'s `ADVANCE` / `WAIT` / `RETRY` / `REPAIR` / `IDLE` — and
  records the retirement under **Limits** so a reader who remembers the old
  vocabulary lands somewhere.
- **`PENDING_TTL_HOURS` is treated as gone.** It appears in no script or
  workflow (`rcc-decided.sh`'s header says selection now reads records, not
  statuses), so the leaf says a commit without a *record* is undecided and a
  `pending` status is safe to disbelieve. `scripts/EACH.md` still names
  `PENDING_TTL_HOURS` in its §3 table and its failure-mode table; that looks
  stale, and it belongs to the `operations/ci/per-commit/` leaf to fix.
- **The failure-class table is a recognition aid, not the repair playbook.**
  Its rows come from `series-check.sh`'s `classify()` and stage 2 of
  `.claude/skills/series-loop.md`; the actions column points onward rather than
  restating the fold-and-replay procedure.
- **The manual-recovery snippet gained a caveat.** `VENDORING.md` presented
  `vendor.sh` as recovery without saying what it does to a `{S}-dev`
  branch; the leaf keeps the procedure and states that it produces one commit
  and is a seed, not a catch-up.
- Nothing named in a recovery procedure was found missing: `vendor.sh`,
  `vendor-one.sh`, `series-check.sh`, `snapshot-accept.sh`, `rconfigure.py`,
  `each.yaml`, `rcc-logs.yaml` all exist as described.

## Durability sweep

**Out, replaced by the mechanism:**

- "looked past, twenty commits deep by default" → "as deep as
  `BASE_SCAN_DEPTH` allows". The variable is already the fix the section
  recommends. *(Partly reversed — see below.)*
- "The same bound is written into `series-advance.sh` and `series-check.sh` as
  a literal twenty" → "hard-coded in `series-advance.sh` and `series-check.sh`,
  which read no such variable". The coupling — raising the depth here means
  editing them too — is the durable fact; the literal is not.
  *(Partly reversed — see below.)*
- "`rcc-logs.yaml` sweeps every 30 minutes" and "compare the age of the `rcc`
  branch tip against its thirty-minute schedule" → the workflow's own sweep
  schedule, with the workflow now linked so the cadence is one click away.
  This one stands: the leaf still names the schedule rather than quoting it.
- "The series loop's rule is twelve hours" → "The series loop's playbook sets
  how long to wait before that". The playbook owns the rule and is linked from
  the top of the leaf. *(Reversed — see below.)*

**Out, as timings nobody will re-measure:** "It costs about twenty seconds" →
"It costs seconds"; "rather than fifteen minutes into a build" → "rather than
deep into a full build". Both served the same point — the gate is cheap enough
to run once per commit — which survives without the stopwatch. "four files at
a time" became "several files in parallel", which is also a more accurate
reading of `xargs -P`.

**Out, because the count only restated what followed:** "Three commands answer
…" → "These commands answer …"; "Two causes, with different fixes" → "The
causes, with their different fixes".

**Kept:** the failure-class table in full — it names each class, which is the
durable form. Exit status 3, the `ADVANCE` / `WAIT` / `RETRY` / `REPAIR` /
`IDLE` verdicts, and the store paths are values, not counts. "Both scripts"
(`vendor.sh` and `vendor-one.sh`, named in the same breath), "exactly one
vendor commit" from a re-vendor (the seed-not-catch-up point), the one line of
`pragma_version.cpp` that differs between clones, and the "two stores" of
finding F1 all stay: in each, changing the number would mean changing the
design.

## Operational defaults put back

The sweep went one step too far on three of its removals, and a later commit
restored them. A default *governs behaviour*: a reader sent to open the script
for the number has been sent away for the thing they came for. The mechanism
stays beside the value in each case.

- **The base scan depth is twenty again** — "as deep as `BASE_SCAN_DEPTH`
  allows, which is twenty commits unless the environment says otherwise".
  Verified: `scripts/vendor.sh` and `scripts/vendor-one.sh` both read
  `${BASE_SCAN_DEPTH:-20}`.
- **The hard-coded twenty is named as the same twenty** — "The same twenty is
  hard-coded in `series-advance.sh` and `series-check.sh`, which read no such
  variable, so a branch that needs a raise here needs them changed too." That
  those two scripts hard-code *the same* number is exactly why the coupling
  bites; naming it is what makes the sentence checkable. Verified: `git log -n
  20` and an `-ge 20` bound in both scripts, neither of which mentions
  `BASE_SCAN_DEPTH`.
- **The playbook's wait is twelve hours again** — "The series loop's playbook
  waits twelve hours before that", with the playbook still linked as the owner
  of the rule.

The `rcc-logs.yaml` cadence was *not* put back: it is a workflow schedule one
click away in a file the leaf links, not a knob the reader is being asked to
reason about.

---

### DIFF
--- a/handbook/operations/vendoring/troubleshooting/README.md
+++ b/handbook/operations/vendoring/troubleshooting/README.md
@@ -1,21 +1,244 @@
 # Troubleshooting
 
-*Stub — this leaf will own its topic;
-today it routes to where the knowledge lives.
-The writing protocol is in [`meta/handbook/`](../../../meta/handbook/);
-the last section holds this leaf's parameters.*
+When a vendoring run goes red:
+how to tell the failure modes apart, what causes each one, and what it needs.
+The repair procedures themselves are playbooks the series loop runs
+([`series-loop/`](/handbook/operations/vendoring/series-loop/README.md));
+what follows is how to reach the right one.
 
-Scope: when a run is red —
-the glue gate, base scans, dropped patches,
-stuck shards, stale snapshots.
+## Where to look first
 
-Today:
+[`scripts/series-check.sh`](/scripts/series-check.sh) is the read-only diagnosis.
+For every series it walks `<S>-green..<S>-dev`,
+looks each commit up in the harvest on branch `rcc`,
+and prints one verdict —
+`ADVANCE`, `WAIT`, `RETRY <sha>`, `REPAIR <sha>`, or `IDLE` —
+beside the in-flight and buffered counts.
+It takes no arguments, discovers the series from the refs,
+and changes nothing, so it is always safe to run first.
 
-* [`scripts/VENDORING.md`](../../../../scripts/VENDORING.md)
-* [`scripts/EACH.md`](../../../../scripts/EACH.md)
+The harvest it reads lives on the orphan branch `rcc`:
+one record per commit at `runs2.d/<xx>/<sha>.ndjson`,
+a failing commit's whole log at `logs2/<sha>.log`,
+and `runs2.ndjson` holding the same records concatenated
+for commits that predate the per-commit layout.
+A matrix leg publishes its record within seconds of deciding a commit,
+and `rcc-logs.yaml` sweeps every 30 minutes for whatever a leg could not publish.
+The record is what decides whether a commit counts as judged;
+the `rcc` commit status is a display surface on the commit list.
+The store's shape and its writers belong to
+[`ci/per-commit/`](/handbook/operations/ci/per-commit/README.md).
 
-To write this leaf:
+Three commands answer "what is vendored here, and when did it last move":
 
-* absorb: the troubleshooting section of `scripts/VENDORING.md`;
-  glue gate, base scans, dropped patches;
-  stuck shards belong to `ci/per-commit/` — link
+```bash
+grep duckdb_version R/version.R            # the DuckDB version string
+git log -1 --grep="^vendor:" --format=%s   # the upstream commit it came from
+git log --oneline --grep="^vendor:" -10    # the last ten vendor commits
+```
+
+## The script refuses to start
+
+Both [`vendor.sh`](/scripts/vendor.sh) and
+[`vendor-one.sh`](/scripts/vendor-one.sh) abort with
+`Error: working directory not clean` before touching anything:
+`git status --porcelain` must be empty.
+Commit or stash first.
+A dirty *upstream* clone is only a warning.
+`vendor-one.sh` additionally refuses unless the working directory
+is the root of its own git worktree.
+
+## The base scan comes up empty
+
+```
+Error: no duckdb/duckdb@ subject among the newest N src/duckdb commit(s)
+```
+
+Both scripts recover where the branch stands in upstream history
+by scanning back over recent commits that touched `src/duckdb/`
+and taking the first `duckdb/duckdb@<sha>` they find in a subject.
+The pathspec narrows the walk and the subject decides,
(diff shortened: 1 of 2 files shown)


krlmlr:
`main` now carries a written leaf for vendoring troubleshooting at reference depth, and it grew a second half on repairing a commit upstream broke (#2493). This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Salvaged from here: that `BASE_SCAN_DEPTH` is honoured by the two vendor scripts only, while `scripts/series-advance.sh` and `scripts/series-check.sh` hard-code the same depth and read no such variable — so raising it for one raises it for neither of the others; and that an empty `Files:` list beside `Error: could not derive glue compile flags (R CMD SHLIB -n)` means the flags could not be derived at all, not that every glue file broke.

Not salvaged: the wedged `pending` finding. `operations/ci/per-commit/` now states the durable form of it directly — the status is a display artifact, selection reads records, and a commit without a record is undecided — which is better than naming a constant that survives only in a plan document.

---

---

PR #2473, opened 2026-08-02
docs(handbook): Write the vendoring model leaf

krlmlr:
## What this leaf now owns

`handbook/operations/vendoring/model/` explains why this package carries a complete copy of the DuckDB C++ engine and why that copy advances one upstream commit at a time. It states the four dev-branch properties with their original numbering preserved (linear, one upstream commit per vendor commit, green per commit, auditable R-side delta), what the invariant buys — bisectability across the engine boundary, where a `git bisect` step lands on one vendor commit naming one upstream commit — and what it costs: one build per vendorable commit with no shortcut for a backlog, a walk that stops at a broken commit rather than patching over it, a branch that cannot be merged into or re-pointed, and glue fixes that lose their own commit. It also owns the vendor commit subject as the only record of where a branch stands in upstream history, and the fork-point rule together with the `main-dev` failure that motivates it. The engine itself, the series-wide guarantees, the scripts, the routine, and red runs are linked, not absorbed.

## What moved

* `scripts/VENDORING.md` — lost §§ "What is Vendoring?", "Why Vendor DuckDB?", "Dev Branch Invariants", "Understanding Vendor Commits", and the rationale, rule, and fork-point definition from "Starting a New Dev Line". It keeps all its mechanics; each absorbed heading is retained with a one-line pointer to the leaf, so the inbound anchors elsewhere in the file still resolve. Its intro sentence no longer claims to cover the invariants. The file is not deleted — the absorption is partial, with `pipeline/` and `troubleshooting/` splitting the rest in parallel.
* `AGENTS.md` — gained a "Where to look" row routing the model to the leaf, and its Vendoring section's restated "one commit at a time" bullet became a pointer instead of a second copy of the invariant. The mechanics row was narrowed to "scripts, troubleshooting".
* `BRANCHES.md` — its §Vendoring pointer to `scripts/VENDORING.md#dev-branch-invariants` now names the leaf, since that anchor no longer holds the content.
* `scripts/README.md` and `.claude/skills/docs-consistency/docs-readme.R` — **unchanged, deliberately.** `VENDORING.md` maps to the internal node `handbook/operations/vendoring`, and with three leaves splitting the file that parent node is still the correct owner: it routes to all three. Remapping it to any single leaf would be wrong, so the `globs =` groups stayed as they are and no regeneration was needed (`docs-readme.R --check` reports the index current). Per the wave convention, no italic backreference line was added to `VENDORING.md` — the generated index carries it.

## Assumptions

* **The fork-point boundary.** I read the *rule* as model and the *procedure* as mechanics. The leaf takes why re-pointing a dev branch breaks invariant 2, the `main-dev` narrative, the rule itself, and the definition of the fork point (newest commit on the first-parent chain of both branches, not `git merge-base`). `VENDORING.md` keeps the `git rev-list`/`awk` snippet that computes it and the day-one steps, since those are the procedure, and the seeding procedure is `series-loop/`'s.
* **The numbered invariants stay whole here, including 1 and 4.** `branches/invariants/` owns the C-numbered series guarantees from `BRANCHES.md`; this list is `VENDORING.md`'s own, cited by number from `VENDORING.md` itself ("invariant 3") and previously from `BRANCHES.md`, so splitting it would break those citations. The leaf's Boundaries section says explicitly that these four say what a vendor commit must be, not what a series promises, and links `branches/invariants/`.
* **Two headings collapsed into one pointer.** "What is Vendoring?" and "Why Vendor DuckDB?" were adjacent and both fully absorbed, so `VENDORING.md` keeps the first heading with a single pointer rather than two consecutive pointer lines. Nothing links to `#why-vendor-duckdb`.
* **`AGENTS.md` got no italic H1 backreference.** Its own "Where to look" table is its backreference mechanism, and several leaves serve the file; a line naming only this leaf would be misleading. A reviewer may prefer the italic line once the other leaves land.
* **The `main-dev` figures are gone** (see the durability sweep below). They came from `scripts/VENDORING.md` and could not be re-derived here, since the `-dev` branches live in the fork and are not reachable from this checkout. Everything else is verified against source: the commit subject formats against `scripts/vendor.sh` and `scripts/vendor-one.sh`, the SHA parsing against `vendor-one.sh`, `series-advance.sh`, and `series-port.sh`, the `DUCKDB_SOURCE_ID` mismatch refusal against `configure`, and per-commit green against `.github/workflows/each.yaml`.
* **`DUCKDB_R_USE_SYSTEM_LIB=1` is framed as a boundary, not a feature of this leaf** — a developer fast path, not a second distribution model — and linked to `build/fast-paths/`. Per the wave rules no build was run under it.
* Pre-existing `../` links in the files I edited were left alone; the separate sweep PR converts them. Links I wrote are root-absolute where they leave their directory, relative where they descend (`AGENTS.md` and `BRANCHES.md` sit at the repository root).

## Durability sweep

**Out, reduced to the qualitative claim the argument needs:** the two `main-dev`
figures — "a `main` commit two weeks older than `v1.5.0`" and "101 first-parent
commits past the fork point", plus the "hundred-odd upstream commits" that
echoed it. They were carried over from `scripts/VENDORING.md` and were the one
thing in the leaf that could not be verified from this checkout, so they were
both unverifiable and stale-prone. The narrative now says the vendored sources
moved backwards to "a `main` commit predating the release" while landing
"everything `main` had accumulated past the fork point, in a single step" —
which is the point: the batch was large enough that a bisect across it answers
nothing.

**Out, because nothing cites it:** "The price is paid in four places" →
"in several places". The bullets are not referenced by number.

**Kept, deliberately:** the four numbered invariants and both mentions of
"four properties". The numbers are identifiers — the leaf itself says
"invariant 2", "invariant 3", "invariant 4", and `scripts/VENDORING.md` points
here saying it keeps this numbering. The set is the design: a fifth property
would be a change to the model, not drift. Also kept: `v1.5.0` and
`v1.4-andium`-style names (which version, not how many), and
"A backlog of *n* commits costs *n* builds", which is a relation rather than
a measurement.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,7 +14,8 @@ start there rather than searching.
 | Build, test, and validate the package locally | this file (below) |
 | Branch model, package flavors, series invariants | `BRANCHES.md` |
 | Release process, modelled as a state machine | `RELEASE.md` |
-| Vendoring mechanics: scripts, invariants, troubleshooting | `scripts/VENDORING.md` |
+| Why the engine is vendored, and the one-commit-per-upstream-commit invariant | [`handbook/operations/vendoring/model/`](handbook/operations/vendoring/model/README.md) |
+| Vendoring mechanics: scripts, troubleshooting | `scripts/VENDORING.md` |
 | Per-commit CI (sharded matrix): design and limits | `scripts/EACH.md` |
 | Operating the vendoring loop (routine playbooks) | `.claude/skills/`: `series-loop.md`, `series-forward.md`, `series-rebase.md`, `series-open.md` |
 | Designs, plans, and historical documents | `plan/README.md`, which names each by path |
@@ -94,7 +95,7 @@ The duckdb-r package vendors (includes a copy of) the DuckDB C++ core library. K
 
 - **Automated Process**: Runs daily via GitHub Actions in `krlmlr/duckdb-r`, gated on the per-commit `rcc` build status
 - **Branch Strategy**: one dev branch per upstream branch (`main-dev` ← `main`, `v1.5-variegata-dev` ← `v1.5-variegata`, `v1.4-andium-dev` ← `v1.4-andium`); see [BRANCHES.md](BRANCHES.md)
-- **One commit at a time**: each vendor commit corresponds to exactly one upstream commit, and must build and pass tests on its own — fold any required glue fix into that commit rather than adding a follow-up
+- **One commit at a time**: the vendoring model — why the engine is vendored, and the invariant every vendor commit satisfies — is [`handbook/operations/vendoring/model/`](handbook/operations/vendoring/model/README.md)
 - **Never modify `src/duckdb/` directly** - changes will be overwritten by vendoring
 - **Patching**: Add files to the `patch/` directory to apply R-specific modifications to vendored code. Send patches upstream as pull requests every once in a while. A patch that no longer applies is deleted by the next vendor run.
 - **Manual vendoring**: Use `scripts/vendor.sh /path/to/duckdb/repo` for testing
--- a/BRANCHES.md
+++ b/BRANCHES.md
@@ -599,7 +599,7 @@ a **linear** first-parent history without merge commits,
 **one upstream commit per vendor commit** forming a contiguous first-parent walk of the tracked upstream branch,
 and a **green** build for every commit.
 These are stated as invariants, with the reasoning and the failure modes, in
-[Dev Branch Invariants](scripts/VENDORING.md#dev-branch-invariants).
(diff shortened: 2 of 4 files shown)


krlmlr:
`main` now carries a written `operations/vendoring/model/` leaf (#2493) — why the engine is vendored rather than submoduled, and the one-commit-per-upstream-commit rule. This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Nothing was salvaged from here. The model is the part of vendoring that is pure *why*, so there was little to verify against the code, and the mechanics it restated belong to `pipeline/` and `series-loop/`.

---

---

PR #2474, opened 2026-08-02
docs(handbook): Write the operations/releases/cran leaf

krlmlr:
## Documentation only — the code moved to #2492

This branch was rebuilt to carry the three `docs(handbook)` commits and
nothing else. `fix(patch): Fix the re2 and mbedtls warnings instead of hiding
them` — everything under `patch/` and `src/duckdb/`, including the careful
account of the re2 fixes, the `patch/0016` split, and the three options for
the `-Wvla` hunk — is now in **#2492**.

**#2492 merges first.** The leaf below records those fixes as done and cites
`patch/0003-Fix-clang-warnings-in-re2.patch` by its new name. Merged the
other way round, it describes a state the tree is not in.

The two branches together reproduce this PR's pre-split tree byte-for-byte;
checked with `git merge-tree --write-tree` against the original tip.

## What this leaf now owns

`handbook/operations/releases/cran/` owns the CRAN-facing half of a release:
how the source tarball is obtained (built from `duckdb/duckdb@main` so the
embedded revision ids are right) and checked before submission — WinBuilder
against R-devel, the CRAN checks page, R-hub via a push to a `cran-*` branch —
what the submission form actually carries, what `cran-comments.md` is for and
how it is kept, and the CRAN policy constraints the source tree lives under
year-round: no warning suppression, the permanent package-size NOTE, and what
`.Rbuildignore` keeps out of the tarball. Boundaries are linked, not repeated:
the state machine to `operations/releases/process/`, the C++ source convention
to `architecture/glue/`, the CRAN test guard to `testing/guards/`, storage
locations to `usage/storage/`, and reverse-dependency checks to
`testing/revdep/`.

## The pragma finding, recorded here and fixed in #2492

Writing the leaf turned up a defect: `tools:::.check_pragmas` greps `src/`
and `inst/include/` for `^\s*#pragma (GCC|clang) diagnostic ignored` — single
spaces, and no space between `#` and `pragma`. Two patches in the stack wrote
their suppressions just outside that pattern, as `#  pragma` and as
`#pragma  GCC  diagnostic  ignored`. The compiler honours both exactly as if
written normally; `R CMD check` sees neither. The stated policy —
`AGENTS.md`, and now this leaf — is that warnings get fixed, not silenced.

**#2492 resolves most of it**, with the evidence: `patch/0003` renamed and
fixed at the source (re2's nested anonymous types and destructor spelling),
`patch/0016`'s `-Wredundant-decls` hunk removed outright, and its `-Wvla`
hunk kept, deliberately, in the invisible spelling. That last one **needs a
maintainer decision** and #2492 lays out three options; this leaf documents
whichever holds, and today says plainly that the spelling is why `R CMD
check` stays quiet.

For contrast the leaf also records the two patches that already honour the
rule: `patch/0008-Avoid-pragma-for-zstd.patch` deletes a suppression
outright, and `patch/0033-clang-macos.patch` fixes the root cause of
`-Wdeprecated-declarations` in fmt.

## What moved

* `RELEASE.md` — lost the whole `### WinBuilder + final checks` section and
  the CRAN-submission paragraph of `### 5 PUBLISHED`, each replaced by a
  one-line pointer to the leaf; gained the visible italic handbook
  backreference under its H1. The rest of the file is untouched — the state
  machine, versioning, and revdep material belong to sibling leaves being
  written in parallel.

`cran-comments.md` is deliberately **not** changed. It briefly gained a
backreference and it was reverted: that file is outbound correspondence, not
documentation of a topic, and its whole text is uploaded verbatim as the
submission comment. The backreference convention exists so a reader standing
in the source tree finds the leaf explaining what they are looking at; nobody
browses `cran-comments.md` for that, and its actual audience is a CRAN
maintainer who should not be reading our internal routing. An HTML comment
would be no better — CRAN receives the raw text. The leaf links to it one-way
and records why that asymmetry is correct rather than an orphan.

## Assumptions

* **`RELEASE.md` is deliberately not deleted.** This is a partial absorption:
  two of the three leaves splitting it are on sibling branches. Conflicts in
  `RELEASE.md` on merge are expected.
* **`r-package-source` is read as an upstream artifact.** `RELEASE.md` calls
  it "the CI build artifact"; no workflow in this repository produces a
  source tarball, and the preflight bullet points at `duckdb/duckdb`'s
  `R.yml`, so the leaf says "the upstream R workflow's `r-package-source`
  artifact". Worth confirming.
* **The maintainer is named by role, not by person.** `RELEASE.md` said "the
  maintainer (currently Hannes) confirms the upload", but `DESCRIPTION`
  currently lists Kirill Müller as `cre` and CRAN mails the confirmation to
  the `Maintainer:` address of the submitted tarball. The leaf therefore says
  "the `cre` entry in `DESCRIPTION`" and drops the name. Whether the old
  wording was stale or described a real hand-off is being confirmed with the
  maintainer separately.
* **The tarball inventory was computed, not assumed** — every top-level entry
  matched against `.Rbuildignore` in R rather than by running `R CMD build`
  (the wave is documentation-only, no builds). `.gitignore` and
  `.Rbuildignore` survive that filter and are listed as surviving; `R CMD
  build`'s own built-in exclusions were not modelled.
* **`RELEASE.md`'s "CRAN rejection" failure-table row was left in place.** It
  restates a fact the leaf now owns, but the table is the process leaf's
  inventory, so it was not edited.
* **The root `README.md` was not touched.** Its `## Documentation` entry
  still points at `RELEASE.md`, which still exists and is still accurate;
  `cran-comments.md` was never listed there.
* **The `Preflight` bullets in `RELEASE.md` were left alone** even though
  they state the same "only the package-size NOTE is tolerated" rule the leaf
  now owns — that section belongs to `operations/releases/process/`.
* **CRAN itself was not reachable** from this environment (the proxy refuses
  `cran.r-project.org`), so the published check results and tarball size were
  not fetched. No claim in the leaf depends on them; the 5 Mb installed-size
  threshold comes from R's own documentation for
  `_R_CHECK_PKG_SIZES_THRESHOLD_`.
* **No durability sweep was run on this leaf.** The numbers it quotes —
  `sizeof(Prog)`, the translation-unit counts, the compiler versions — are
  measurements of a specific investigation rather than inventory of the tree,
  so there was nothing of that class to remove. The investigation itself, and
  the caveat that its compiler evidence is Linux/x86-64 only, is in #2492.

## Verification

`Rscript .claude/skills/docs-consistency/docs-readme.R --check` and the
handbook link check (no dangling target, none starting with `../`) are clean
on this branch. `R CMD check` and a full package build were not run — not
feasible in this environment — and no claim here depends on them.
`tools:::.check_pragmas(".")` returns `character(0)` both before and after
#2492, so neither PR changes what that check reports today.

---

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,7 @@
 # R Package Release Process
 
+*Handbook: [`operations/releases/cran/`](/handbook/operations/releases/cran/README.md).*
+
 This document describes the release process for one DuckDB release line as a
 **finite state machine (FSM)**. It is the operational companion to
 [`BRANCHES.md`](BRANCHES.md), which defines the branch model, the package
@@ -219,12 +221,7 @@ git rebase origin/L && git push origin L-lts
 
 ### WinBuilder + final checks
 
-Obtain the source tarball from the CI build artifact (`r-package-source`) or the
-post-CI GitHub release (`duckdb_<version>.tar.gz`). Build it from
-`duckdb/duckdb@main` rather than a fork, so the git revision ids used to fetch
-extensions are correct. Upload to WinBuilder (R-devel):
-<https://win-builder.r-project.org/upload.aspx>. Apart from the known
-package-size NOTE, warnings and notes are blockers.
+See [`operations/releases/cran/`](/handbook/operations/releases/cran/README.md).
 
 ### 5 PUBLISHED — tag & publish
 
@@ -235,12 +232,10 @@ git tag vX.Y.Z origin/L-lts   # or origin/L per series convention
 git push origin vX.Y.Z
 ```
 
-For the **current** line (the CRAN `duckdb` package), submit the source tarball at
-<https://cran.r-project.org/submit.html>. The maintainer (currently Hannes)
-confirms the upload, after which CRAN's automated checks run. CRAN acceptance is
-**asynchronous** — it can take days and overlaps RESET and the next TRACK. If CRAN
-rejects, fix on `stable` and re-enter CUT as a follow-up patch. LTS lines publish
-to r-universe only and have no CRAN tail.
+For the **current** line (the CRAN `duckdb` package), the submission itself —
+and the asynchronous acceptance tail that overlaps RESET and the next TRACK —
+is [`operations/releases/cran/`](/handbook/operations/releases/cran/README.md).
+LTS lines publish to r-universe only and have no CRAN tail.
 
 ## RESET — re-baseline
 
--- a/cran-comments.md
+++ b/cran-comments.md
@@ -1,5 +1,7 @@
 duckdb 1.5.5
 
+*Handbook: [`operations/releases/cran/`](/handbook/operations/releases/cran/README.md).*
+
 ## Cran Repository Policy
 
 - [x] Reviewed CRP last edited 2026-05-31.
--- a/cran-comments.md
+++ b/cran-comments.md
@@ -1,7 +1,5 @@
 duckdb 1.5.5
 
-*Handbook: [`operations/releases/cran/`](/handbook/operations/releases/cran/README.md).*
-
 ## Cran Repository Policy
 
 - [x] Reviewed CRP last edited 2026-05-31.
--- a/handbook/operations/releases/cran/README.md
+++ b/handbook/operations/releases/cran/README.md
@@ -81,6 +81,16 @@ file — version, timestamp, and commit SHA of a submitted release —
 that `devtools` writes after a successful upload.
 No such file is in the tree.
 
+The link to `cran-comments.md` is deliberately one-way:
+it carries no backreference to here,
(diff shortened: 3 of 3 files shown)


krlmlr:
`main` now carries a written leaf for CRAN at core depth (#2493), covering the submission sequence and the policy the tree obeys year-round. This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Salvaged from here: that CI turns the installed-size check off. `custom/before-install/action.yml` exports `_R_CHECK_PKG_SIZES_=FALSE`, so the standing package-size NOTE cannot fail a matrix run — and neither can a size regression, which is first seen at submission time.

Not salvaged: the tarball inclusion and exclusion lists, which `.Rbuildignore` already is, and the pragma-respelling account, which the glue leaf now owns.

---

---

PR #2475, opened 2026-08-02
docs(handbook): Write the branches/model leaf

krlmlr:
## What this leaf now owns

`handbook/branches/model/` now explains the series and their refs: what a series is and why it is discovered from its refs rather than configured, which of the two repositories holds what, the four refs a series carries (`{S}-build`, `{S}-dev`, `{S}-green`, `{S}-build-base`) with what each means and how far each may move, forward counterparts and why a cutover is a human's call, the source-of-truth rule with the forward-port chain and how it is kept, and the release-cycle mapping — mid-cycle, pre-release, feature freeze, a new minor, a patch release, and LTS expiry. Facts belonging to siblings are linked, not repeated: flavors, the numbered invariants, version counters, the pipeline, the loop, and the release state machine.

(Refs are written `{S}-build` here. The leaf itself uses angle brackets; GitHub deletes those from a pull-request body even inside backticks.)

## What moved

* `BRANCHES.md` — lost §§ "Overview", "Repositories", "Branch Overview" (with "Branch series structure"), "Source of Truth" (with "Keeping derived branches in sync with main"), and "Release Cycle Mapping" (with its phase subsections). Each heading is **kept**, carrying a one-line pointer to the leaf, so the `#source-of-truth`, `#release-cycle-mapping` and `#on-patch-release-v145` anchors that `scripts/VENDORING.md` and `BRANCHES.md`'s own § "Synchronization" link still resolve. Gained the visible handbook backreference under its H1. The rest of the file — package components, flavors, series invariants, synchronization/vendoring, patch stack, version numbering, tooling — belongs to leaves still being written and is untouched.
* `README.md` — the "See `BRANCHES.md` for the full model" line under § Flavors now points at the leaf, and the § Documentation entry now names `handbook/branches/` (with the leaf named inside it) instead of `BRANCHES.md`, which it still mentions as the not-yet-moved remainder.

## Assumptions

* **"Replace each absorbed heading with a one-line pointer" read as: keep the heading, replace its body.** Deleting the headings outright would have broken five inbound anchor links from `scripts/VENDORING.md` and from a `BRANCHES.md` section owned by another leaf, and fixing those would have meant editing another leaf's text. Confirm this is the intended shape; if not, the headings can be dropped and the inbound anchors retargeted in a follow-up.
* **The "Ongoing" forward-port chain in `BRANCHES.md` § Synchronization was left in place** even though the leaf now owns that chain. § Synchronization belongs to `operations/vendoring/pipeline/`, so its owner should retire that paragraph rather than this PR reaching into it.
* **`BRANCHES.md` § Synchronization says `{S}-dev` is consumed "up to 25 commits at a time"; the default in `scripts/series-advance.sh` is a hundred.** The leaf now states the hundred and records that `BRANCHES.md` long said twenty-five, which was never the default — see "The chunk size, restated" below. The stale sentence in `BRANCHES.md` itself sits outside this leaf's absorption and is left for § Synchronization's owner. (Reported independently by the agent writing `operations/vendoring/series-loop/`.)
* **The verified state of the refs is stated as of today, in a "What exists today" subsection.** The loop refs exist for `main`, `v1.5-variegata` and `v1.4-andium`; `main` has a live `main-fwd` counterpart; the legacy `{S}-dev-base` refs still exist but no script or workflow in the tree reads them — only the release FSM in `RELEASE.md` still models them. The leaf says so plainly rather than presenting either layout as the whole truth. Checked with `git ls-remote` against `krlmlr/duckdb-r` and the GitHub API against `duckdb/duckdb-r`.
* **The loop's cadence is stated as not visible in the repository.** Nothing in `.github/workflows/` fires the series loop or `vendor-one.sh`; it is a scheduled Claude routine configured elsewhere. The leaf says that rather than repeating a cadence it cannot check.
* **The three proposed sync mechanisms are recorded as declined, not merely absent.** `scripts/sync-to-derived.sh`, a stale-branch workflow and a `needs-forward-port` label do not exist in the tree, and `scripts/series-port.sh` covers what they were for. If any of them is still wanted, that verdict needs correcting.
* **r-universe's package-to-branch mapping is stated as unverifiable from here** (it lives in the registry), so the leaf says `{S}-green` is what *should* be built, matching `.claude/skills/series-loop.md`, and points at `branches/flavors/` for the intended mapping.
* No issues were listed for this leaf, so none were drained. No build was run; every claim is from the refs, the scripts and the workflows.

## Durability sweep

**Out.** The counts in "What exists today": "three base series", "all four refs",
"the two parked baselines". The series and the refs are still named, which is
what a reader came for; opening a series no longer falsifies a sentence.

The sweep also removed the chunk size, on the grounds that it is a default in a
script. That one was put back — see below.

**Stayed.** The four refs as a design — the section heading, the table, and
"all four exist from a series' first day, equal at its seed tip". Changing that
number would mean changing the model, so it is a fact and not a hostage. The
`{S}-dev-base` paragraph in full: it is an observation framed as one and it
carries the argument that a ref nothing reads is still a live boundary between
two layouts. The `TOOLING / MIXED / OTHER / VENDOR` classes in
`series-port.sh`, likewise the design. The upstream cadence and LTS window,
which are upstream policy. The workflow schedules in § Limits, which name
which workflow does what rather than counting anything.

## The chunk size, restated

The sweep replaced "at most 100 commits per pass" with a pointer to the
script's default. A later commit reversed that: a default that governs
behaviour is a fact worth stating, the same way each build knob's default is
stated in `build/configuration/`. It bounds how much a single pass can move,
which is how a reader reasons about the loop.

The leaf now says that `scripts/series-advance.sh` appends at most one chunk of
`{S}-build` commits to `{S}-dev` per pass, **a hundred by default, overridable
per call**, and says what the bound buys: it keeps a pass reviewable and keeps a
red run from stranding an unbounded backlog. It also records that `BRANCHES.md`
long said twenty-five, which was never the default, so the wrong number does
not come back.

Verified against the script: `series-advance.sh` documents
`series-advance.sh {series} [chunk-size]` with "chunk default 100" and takes
`chunk=${2:-100}`.

---

### DIFF
--- a/BRANCHES.md
+++ b/BRANCHES.md
@@ -1,5 +1,7 @@
 # Branching Strategy
 
+*Handbook: [`branches/model/`](/handbook/branches/model/README.md).*
+
 Version numbers are given at the time of writing (March 2026) and may be outdated by the time you read this.
 The branching strategy is expected to remain stable.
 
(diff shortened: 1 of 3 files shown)


krlmlr:
`main` now carries a written `branches/model/` leaf (#2493), naming the four refs of a series and the fork they live in. This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Nothing was salvaged from here. Its limits section came closest — that the r-universe branch-to-package mapping lives outside this repository, and that nothing in `.github/workflows/` fires the series loop — but the flavors leaf already carries the first as its deepen line, and the per-commit subtree already states that nothing schedules `each-rcc`.

---

---

Issue #2425, opened 2026-07-28
Test failure on Windows ARM64


jeroen:
See https://github.com/r-universe/duckdb/actions/runs/30353588565/job/90262612127

eitsupi:
My understanding is that this is because DuckDB does not distribute extensions for arm64 MinGW, and this may be resolved in DuckDB v2, which is scheduled to migrate to C extension API.

In the meantime, I think the problem can be solved by specifying an extension that already uses the C API instead of the ICU extension.

krlmlr:
Another example: https://github.com/r-universe/duckdb/actions/runs/30533304494/job/90847736895 .

@eitsupi: Do you have an example for such an extension?

eitsupi:
A quick search under the duckdb organization suggests that only odbc-scanner is relevant.

However, it would obviously be better to ask someone at DuckLab about this.

krlmlr:
@staticlibs: Will there be support for Windows ARM64 MinGW extensions?

jeroen:
The [documentation](https://duckdb.org/docs/current/extensions/overview) says:
 
> All operating systems are supported for both the AMD64 and the ARM64 architectures.

So maybe it's just a matter of getting the URL right?

eitsupi:
Windows versions other than Go and R are MSVC targets, and their extensions are built.
amd64 MinGW is also built (though it is often skipped).

arm64 MinGW is not built at all because it does not exist as a target.

https://github.com/duckdb/extension-ci-tools/blob/64aec33f4f659eed8e542507b3afa3b262b37e42/config/distribution_matrix.json#L60-L87

staticlibs:
@krlmlr 

Hi! @eitsupi is correct, currently extensions are not built for arm64 MinGW. It was considered in duckdb/duckdb#21727 but was not moved forward. Currently there is no timeline for adding arm64 MinGW extensions.

krlmlr:
Verified by experiment:

https://github.com/krlmlr/duckdb-r/actions/runs/30927023956/job/92051868062#step:8:69

(Repo to be renamed, link may go stale.)

> The file was built for the platform 'windows_arm64',
> but we can only load extensions built for platform 'windows_arm64_mingw'.

Action to take:

- Document limitation
- Add a canary test under Windows ARM that checks that installing a C++ extension is _failing_
- Test that installing a C API extension (odbc-scanner?) works, and move the test to that extension

Scheduling the C API check now.

[...]

PR #2476, opened 2026-08-02
docs(handbook): Write the testing/suite leaf

krlmlr:
## What this leaf now owns

[`handbook/testing/suite/`](https://github.com/duckdb/duckdb-r/blob/claude/handbook-testing-suite-zayxqn/handbook/testing/suite/README.md)
explains the `tests/` layout as it actually is — `tests/testthat.R` as the
`R CMD check` entry point, the flat directory of `test-*.R` files and the three
naming conventions they follow, the `helper-*.R` files and what each one
sets up, `setup.R`, the `data/` Parquet fixtures, and the fact that
the test fixtures themselves (`local_con()`) are package code in
`R/test-fixtures.R` rather than helpers. It then covers running the suite,
running one file (`filter` is a regex, anchor it), the fast edit-test loop, and
the manual scenario to drive by hand after a glue change. Boundaries link out
rather than repeat: the mechanism behind `DUCKDB_R_USE_SYSTEM_LIB` is
`build/fast-paths/`'s, snapshots are `testing/snapshots/`'s, the CRAN and
flavor guards are `testing/guards/`'s, reverse dependencies are
`testing/revdep/`'s, and what CI runs is `operations/ci/`'s.

## What moved

* `AGENTS.md` — lost the bodies of §§ "Run Tests", "Manual Validation",
  "Running Tests", "Manual Testing Scripts" and "Test Development"; each heading
  is kept with a one-line pointer to the leaf. Gained the visible italic
  handbook backreference under its H1. The file is not deleted — it shrinks to
  routing.
* `tests/testthat.R` — gained a plain `# Handbook: handbook/testing/suite/`
  comment below its header block, so someone standing in `tests/` finds the leaf.

## Assumptions

* **`## Validation` in `AGENTS.md` was left alone**, although its first bullet
  restates the manual end-to-end scenario I absorbed. It is not one of the five
  sections this leaf was assigned, and five sibling leaves are editing the same
  file in parallel; whoever owns it should fold the duplicate away. Its claim
  that "one test failure is expected in clock function tests" looks stale — the
  clock translations live in `test-backend-dbplyr__duckdb_connection.R` and skip
  when `clock` is not installed — but I did not verify it by running anything.
* **The `Where to look` routing table in `AGENTS.md` still says this file owns
  "Build, test, and validate the package locally".** Left as is to avoid a
  three-way conflict with the sibling leaves; it wants one update once they land.
* **The root `README.md` was not touched.** Its `## Documentation` entry
  describes `AGENTS.md` as covering "build, test, and where to look for
  everything else", which the split makes partly stale — but `AGENTS.md` is not
  deleted, so the entry is not dangling, and the same three-way conflict applies.
* ~~**Counts are stated where they are stable and hedged where they are not**:
  "close to seventy" test files (68 today), "fifteen files" of glue in `src/`
  (verified), three Parquet fixtures (verified), eight `test-storage-*.R` files
  (verified). The "about 45 seconds" for a full run is carried over from
  `AGENTS.md` unverified — no build was run for this wave.~~ Superseded by the
  durability sweep below: every one of these is now out of the leaf.
* **#2425 is answered as a boundary, not closed.** The leaf states today's
  behavior — `test-duckdb-extensions.R` asserts `INSTALL icu` succeeds wherever
  extensions are supported, and there is no prebuilt ICU for arm64 MinGW — and
  attributes the diagnosis to the issue thread (I could not read the r-universe
  job log). The verdict is a code fix in the test, so the issue stays open.
* **`setup.R`'s effect** is described as pointing arrow's DuckDB integration at
  the suite's own connection. That reads the `arrow_duck_con` option as arrow's
  (the name and `test-arrow.R`'s comment about `arrow_duck_connection()` both
  say so); nothing in this repo consumes it.

## Durability sweep

A targeted pass for one defect class: facts a routine commit invalidates. This
leaf was the wave's worst offender, and its own assumptions bullet said so.

**Removed.**

* "close to seventy of them" test files — cut. The flat directory and the
  `test-{topic}.R` naming convention are the durable content: the convention
  tells a reader what to expect from *any* file, which a tally never did.
  (The leaf spells the placeholder with angle brackets; GitHub deletes those
  from a pull-request body even inside backticks, so this description uses
  braces.)
* "three Parquet fixtures" → "the Parquet fixtures". Adding a fixture is a
  routine commit.
* "the three skip predicates" → "the skip predicates". A new test needing a new
  predicate is routine; the list that follows still names each one.
* "`setup.R` is a single line" → "`setup.R` does one thing". The point was
  always that it does one thing, not that it occupies one line.
* "The whole suite, about 45 seconds" — cut outright. The figure came over from
  `AGENTS.md` with no run behind it, and nobody will re-measure it.
* "runs all eight `test-storage-*.R` files" → "runs every `test-storage-*.R`
  file". The argument is that an unanchored filter matches broadly; the count
  only illustrated it.
* "fifteen files" of glue, "a no-op load takes a second, a load after one edited
  glue file about four", and "ten to fifteen minutes" for the engine build —
  reduced to seconds against minutes. The order of magnitude *is* the argument
  ("the difference between a loop and a wait"), and it survives a faster
  machine, a warm ccache, and a changed glue file count. None of the figures
  did.

**Kept.**

* `dbitest_version` `"1.8.3"` — a version number identifying *which* version,
  not how many.
* "A name is one of three things" — this is the naming convention itself, and a
  fourth category would be a deliberate change to the convention rather than a
  routine commit. Load-bearing, so it stays.
* #2425 — an issue number identifies the open bug.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,7 @@
 # DuckDB R Package - Operational Instructions
 
+*Handbook: [`testing/suite/`](/handbook/testing/suite/README.md).*
+
 R package that contains a vendored copy of the DuckDB C++ library and glue code for R, including a DBI and a relational interface.
 
 ## Where to look
@@ -39,23 +41,11 @@ For an interactive edit-build-test loop, prefer the prebuilt-libduckdb fast path
 
 ### Run Tests
 
-- `R -q -e "testthat::test_local()"` -- runs all tests. Takes about 45 seconds. NEVER CANCEL: set timeout to 5+ minutes.
-- `R -q -e "testthat::test_local(filter = '^name$')"` -- runs specific test file by name
+See [`testing/suite/`](/handbook/testing/suite/README.md).
 
 ### Manual Validation
 
-- ALWAYS test basic DuckDB functionality after making changes by running a complete scenario
-- Test connection, table creation, data insertion, and querying:
-
-    ```r
-    library(duckdb)
-    con <- dbConnect(duckdb())
-    dbExecute(con, "CREATE TABLE test (id INTEGER, name VARCHAR)")
-    dbExecute(con, "INSERT INTO test VALUES (1, 'Alice'), (2, 'Bob')")
-    result <- dbGetQuery(con, "SELECT * FROM test ORDER BY id")
-    print(result)
-    dbDisconnect(con, shutdown=TRUE)
-    ```
+See [`testing/suite/`](/handbook/testing/suite/README.md).
 
 ### Format Checking (Optional - Has Known Issues)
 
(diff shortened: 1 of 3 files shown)


krlmlr:
`main` now carries a written `testing/suite/` leaf (#2493), covering the `filter` regex, the layout, and the helper constraints that are not obvious from reading one helper. This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Nothing was salvaged from here. Its additions were rosters — test files, helpers, skip predicates — which `tests/testthat/` already is, and the reasoning behind `skip_on_flavor()` that it wrote out is already the inline header of `helper-skip.R`.

---

---

PR #2471, opened 2026-08-02
docs(handbook): Write the operations/releases/process leaf

krlmlr:
## What this leaf now owns

`handbook/operations/releases/process/` owns the release state machine: the four clusters (TRACK, STABILIZE, CUT, RESET) and the property that separates them, the Mermaid state diagram, the state table with entry conditions and gates, the procedure for each state, multi-line coordination and its single hard barrier, the failure-and-rollback table, and the mapping from cluster to the series invariants it must preserve. It states its own limits: no script or workflow knows about clusters or states, so the FSM is maintainer discipline over CI signals rather than something enforced, and the steps from 3 PROMOTED through 5 PUBLISHED are typed by hand because `scripts/promote-dev.sh` and `scripts/release-lts.sh` exist only as proposed tooling in `BRANCHES.md`.

`RELEASE.md` is being split three ways in parallel, so this absorption is partial by design. Where a state&#39;s steps reach into a sibling&#39;s territory — the 0.2/0.4 reverse-dependency gates, the version bump in 4 MERGED, the CRAN submission in 5 PUBLISHED, the development-version move in RESET — the leaf keeps the step and links `testing/revdep/`, `operations/releases/versioning/`, and `operations/releases/cran/` rather than explaining it.

## What moved

* **`RELEASE.md`** — loses the state machine (clusters, diagram, state table, TRACK, STABILIZE and its preflight/0.1/0.3/0.5 sections, CUT and its 1→2→3 / 4 MERGED / WinBuilder sections, RESET, multi-line coordination, failure and rollback, invariants per cluster). Keeps the three sibling-owned sections — reverse-dependency checks, the `fledge::bump_version()` step, the CRAN submission paragraph, and the RESET version paragraph — each under a one-line pointer to this leaf, plus the `BRANCHES.md` companion sentence. Gains the visible italic handbook backreference under its H1. **Not deleted**: a later cleanup can remove it once the CRAN and versioning splits have landed too.
* **`README.md`** — the `## Documentation` list entry for `RELEASE.md` now names `handbook/operations/releases/` as the release-process area and points at `RELEASE.md` only for what has not moved there yet.

No `scripts/` ownership changed, so `.claude/skills/docs-consistency/docs-readme.R` and the generated `scripts/README.md` are untouched.

## Assumptions

[...]

PR #2477, opened 2026-08-02
docs(handbook): Write the vendoring pipeline leaf

krlmlr:
## Documentation only — the code moved to #2491

This branch was rebuilt to carry the five `docs`/`docs(handbook)` commits and
nothing else. Every change to `scripts/vendor.sh`, `scripts/vendor-one.sh`,
`scripts/rconfigure.py` and the generated `scripts/README.md` is now in
**#2491**, together with the comment strip.

**The split is by artifact, not by intent**, and that is deliberate: the
comment deletions are a documentation change in spirit, but they are
interleaved with the functional hunks in the same three files, and splitting
a file across two PRs would guarantee conflicts between our own branches.
`scripts/VENDORING.md` stays here, being prose that happens to live under
`scripts/`.

**#2491 merges first.** The leaf below describes the *fixed* behaviour —
exit 4, the retirement message, the clean tree after a no-op run — as the
pipeline's rule rather than as a caution. Merged the other way round, it
documents behaviour the scripts do not yet have.

The two branches together reproduce this PR's pre-split tree byte-for-byte;
checked with `git merge-tree --write-tree` against the original tip.

## What this leaf now owns

`handbook/operations/vendoring/pipeline/` explains the machinery that turns
one upstream DuckDB commit into one vendor commit: `vendor.sh` and
`vendor-one.sh` (how each picks its commit, the shared nine-step run,
`--commits`, `VENDOR_REPO`, and the glue gate that exits 3 with the breaking
commit at `HEAD`), what `rconfigure.py` regenerates, the patch-stack
lifecycle — adding, dropping, and breaking a patch in a vendor run — the
vendor commit message and why its subject is machine-readable state, the
`DESCRIPTION` merge driver with its prefix gate and the `setup-git.sh`
wiring, and vendoring by hand including the mandatory `rm -f src/*.o`. It
closes with its limits: `vendor-rfuns.sh` is a separate path, and seeding a
new series is not a plain vendor run. The schedule and playbooks are linked
to `series-loop/`, how far each firing runs the scripts to `branches/model/`,
red runs to `troubleshooting/`, the counters themselves to
`releases/versioning/`, the build knobs to `build/configuration/`, and the
tarball caution to `build/source-build/`.

The rule the leaf now states once and plainly, because #2491's exit-4 change
is what makes it true: **a vendor run either succeeds or dies trying, and
talks about it.** Nothing is dropped, skipped, or worked around to keep a run
alive; every stop names what stopped it, where it left the tree and the
upstream clone, and what to do next.

## What moved

* **`scripts/VENDORING.md`** — lost §§ "The Vendoring Scripts", "Version
  Counters and the Merge Driver", "Manual Vendoring", "Understanding Vendor
  Commits", "Files and Directories" and "Development Guidelines"; each
  heading remains with a one-line pointer at the leaf, so anchors elsewhere
  keep resolving. Two surviving cross-references (the commit-by-commit loop,
  the patch stack) now point at the leaf.
* **`BRANCHES.md`** — lost § "Patch Stack" and the "The DESCRIPTION merge
  driver" subsection of § "Version Numbering", both replaced by pointers;
  gained the visible italic handbook backreference under its H1. Its
  § Synchronization › Vendoring also said `-dev` is consumed "**up to 25
  commits at a time**" — a default that has never been the default:
  `scripts/series-advance.sh` has `chunk=${2:-100}`. Two sibling leaves
  tripped over the contradiction and correctly left it, because that section
  is this leaf's territory. The sentence now states the shape and links
  `branches/model/`, which owns the number. (`RELEASE.md` carries the same
  false 25 at line 111; #2471 deletes that section wholesale, so it is left
  alone here to avoid a conflict. `scripts/EACH.md`'s 25 is an example batch
  size, not a claim about the default — that document already says the cap is
  100.)
* **`README.md`** — the `## Vendoring` section and the `## Documentation`
  entry now point at `handbook/operations/vendoring/` instead of
  `scripts/VENDORING.md`. Its `## Building` section's clone-layout
  instruction was **wrong** and is corrected — see below.
* **`AGENTS.md`** — the sentence describing the two vendor scripts as
  behaving differently around a failing patch is gone, because after #2491
  they do not.
* **`build/source-build/` (#2481)** — the "never run `R CMD build` in a
  working tree you still need" caution moved there, restated in that leaf's
  voice beside the `cleanup` script it is about. Removed here.
* **`build/configuration/` (#2479)** — the `DUCKDB_BUILD_UNITY` row and the
  `DUCKDB_R_BINDIR`/`CFLAGS`/`LIBS` paragraph are removed there, because
  #2491 deletes both knobs. That PR no longer depends on this one, but it
  does depend on #2491 landing.

## The clone layout was wrong everywhere

The leaf, the root `README.md`, and `AGENTS.md`-era prose all said the
`duckdb-r` clone sits "one level deeper" than the `duckdb` clone, e.g.
`duckdb` and `R/duckdb-r`. Read from the code, that is off by one:

* `vendor.sh` runs `cd "$(dirname "$0")"/..` and `vendor-one.sh` runs
  `cd "${VENDOR_REPO:-$(dirname "$0")/..}"`, so both resolve relative paths
  **from the package root**.
* The default source repository in both is
  `upstream_basedir=../../../${project}` — **three** levels up from the
  package root.

So the package clone has to sit two directories *deeper* than the upstream
clone. With the upstream clone at `~/git/duckdb`, `~/git/R/duckdb/duckdb-r`
is a package clone the defaults find with no argument — which is the
maintainer's actual layout. The old root `README.md` block
(`cd R/duckdb-r && scripts/vendor.sh`) could not have worked as written. Both
the leaf and the root `README.md` now state it correctly and use `~/git/...`
paths; any other layout still works by passing the upstream path as the
positional argument.

## Issues linked as boundaries

* **#2489 — the vendored tree is not byte-reproducible across clones.**
  `DUCKDB_SOURCE_ID` is an abbreviated commit id and git sizes the
  abbreviation from the clone it runs in, so one upstream commit vendors as
  `7300522cf0` from one clone and `7300522cf07` from another.
* **#2490 — `rconfigure.py` rewrites all ~3550 files unconditionally**,
  invalidating git's stat cache twice per candidate, including for candidates
  that are skipped.

Both are linked from the leaf in a line each, instead of being narrated as
properties of a diff; both are changes to vendored output rather than to the
scripts' control flow, which is why neither is in #2491. The earlier claim
that the second was "one of the items in
`plan/PLAN-vendoring-simplification.md`" was not true — the plan does not
carry it.

Two further intents are linked rather than restated: **#2488** (should the
merge driver keep the larger version instead of *ours*?) from the prefix-gate
paragraph, and **`duckdb/duckdb#22811`** from the jemalloc sentence, so a
reader sees that fact is provisional and where to check it.

## What was cut as belonging elsewhere

* "the series loop passes 100" → the consumption bound is `branches/model/`'s;
  the leaf links it, and so does the corrected sentence in `BRANCHES.md`.
* "There is no vendoring workflow" → `series-loop/` already states it; the
  duplicate limit is gone, and the surviving `series-loop/` link says that
  leaf also owns what schedules these scripts.
* `VENDOR_REPO`'s two-paragraph rationale → one sentence plus a link to
  `series-loop/`'s first stage.
* The `R CMD build` caution → `build/source-build/` (#2481).

**Judged shared and kept**: the hand-replay section — `rm -f src/*.o`,
ccache, `OBJECTS = $(GLUE) $(SOURCES)` failing fast, and the bimodal
build-time figures. These are what a person driving `vendor-one.sh` by hand
needs; `series-loop/` is about the routine's stages and carries no cost
model.

## Verification

`Rscript .claude/skills/docs-consistency/docs-readme.R --check` and the
handbook link check (no dangling target, none starting with `../`) are clean
on this branch, as are `bash -n` on both vendor scripts and
`python3 -m py_compile scripts/rconfigure.py` — unchanged here, and verified
separately on #2491.

The one thing this leaf asserts that cannot be checked from this branch alone
is that the scripts behave as described; that is #2491's diff, and the merge
order above is what keeps the two honest.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -96,7 +96,7 @@ The duckdb-r package vendors (includes a copy of) the DuckDB C++ core library. K
 - **Branch Strategy**: one dev branch per upstream branch (`main-dev` ← `main`, `v1.5-variegata-dev` ← `v1.5-variegata`, `v1.4-andium-dev` ← `v1.4-andium`); see [BRANCHES.md](BRANCHES.md)
 - **One commit at a time**: each vendor commit corresponds to exactly one upstream commit, and must build and pass tests on its own — fold any required glue fix into that commit rather than adding a follow-up
 - **Never modify `src/duckdb/` directly** - changes will be overwritten by vendoring
-- **Patching**: Add files to the `patch/` directory to apply R-specific modifications to vendored code. Send patches upstream as pull requests every once in a while. A patch that no longer applies is deleted by the next vendor run.
+- **Patching**: Add files to the `patch/` directory to apply R-specific modifications to vendored code. Send patches upstream as pull requests every once in a while. A patch whose change has landed upstream is deleted by the next vendor run; one that has genuinely broken stops `vendor-one.sh` instead.
 - **Manual vendoring**: Use `scripts/vendor.sh /path/to/duckdb/repo` for testing
 - **Full documentation**: See [VENDORING.md](scripts/VENDORING.md) for complete details
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -96,7 +96,7 @@ The duckdb-r package vendors (includes a copy of) the DuckDB C++ core library. K
 - **Branch Strategy**: one dev branch per upstream branch (`main-dev` ← `main`, `v1.5-variegata-dev` ← `v1.5-variegata`, `v1.4-andium-dev` ← `v1.4-andium`); see [BRANCHES.md](BRANCHES.md)
 - **One commit at a time**: each vendor commit corresponds to exactly one upstream commit, and must build and pass tests on its own — fold any required glue fix into that commit rather than adding a follow-up
 - **Never modify `src/duckdb/` directly** - changes will be overwritten by vendoring
-- **Patching**: Add files to the `patch/` directory to apply R-specific modifications to vendored code. Send patches upstream as pull requests every once in a while. A patch whose change has landed upstream is deleted by the next vendor run; one that has genuinely broken stops `vendor-one.sh` instead.
+- **Patching**: Add files to the `patch/` directory to apply R-specific modifications to vendored code. Send patches upstream as pull requests every once in a while. A patch whose change has landed upstream is deleted by the next vendor run; one that has genuinely broken stops the run instead.
 - **Manual vendoring**: Use `scripts/vendor.sh /path/to/duckdb/repo` for testing
 - **Full documentation**: See [VENDORING.md](scripts/VENDORING.md) for complete details
 
--- a/BRANCHES.md
+++ b/BRANCHES.md
@@ -1,5 +1,7 @@
 # Branching Strategy
 
+*Handbook: [`operations/vendoring/pipeline/`](/handbook/operations/vendoring/pipeline/README.md).*
+
 Version numbers are given at the time of writing (March 2026) and may be outdated by the time you read this.
 The branching strategy is expected to remain stable.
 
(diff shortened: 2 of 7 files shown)


krlmlr:
`main` now carries a written leaf for the vendoring pipeline at reference depth (#2493), so this branch predates the tree it edits, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

This branch salvaged more than any other. Three findings from it survive: the clone layout — both scripts `cd` to the package root and default to `../../../duckdb`, so the package must sit two levels deeper than the upstream clone, and `README.md` still says one; that a local commit-by-commit replay must delete `src/*.o` between commits, because `src/include/deps.mk` filters every `duckdb/` path out of the recorded dependencies and `make` therefore never recompiles the glue after vendoring; and that the merge driver's prefix gate declines to renumber a series rebased onto a new patch release, so the vendor counter has to be re-applied by hand (#2488).

All three are recorded in `plan/history/2026-08-handbook-wave-salvage.md` — #2507, [file](https://github.com/duckdb/duckdb-r/blob/claude/handbook-wave-salvage-zayxqn/plan/history/2026-08-handbook-wave-salvage.md) — with the README layout sentence listed as an unfixed defect.

---

---

PR #2478, opened 2026-08-02
docs(handbook): Write the branches/invariants leaf

krlmlr:
## What this leaf now owns

`handbook/branches/invariants/` owns what every series guarantees: the
C-numbered statements — S1–S4, L, A1–A3, F1–F2, V1–V4, G1–G2, C1–C2, P1–P4 —
together with the tree-diff framing they must be read under and the
cost-of-linearity table that justifies accepting the rebases. The value the
page adds over the source is the second half of every entry: what would
actually catch a violation. C1 has a gate of its own (`each.yaml` writes a
verdict per commit, `series-advance.sh` fast-forwards `{S}-green` only over an
unbroken prefix of successes); G2 and the vendor counter in V3 hold
mechanically because `vendor-one.sh` is the only thing that writes them; G1 is
measured and closed by `series-port.sh` as a stage of the series loop. Every
other invariant is a convention, and the page says so in those words, per
entry. The refs themselves stay `branches/model/`'s, the flavor tooling
`branches/flavors/`'s, the per-commit marker `operations/ci/per-commit/`'s,
and the one-commit-per-upstream-commit rule `operations/vendoring/model/`'s.

(Series refs are written `{S}-green` here. The leaf itself uses angle
brackets; GitHub deletes those from a pull-request body even inside backticks.)

## What moved

- **`BRANCHES.md`** — lost § "Series Invariants" in full (the section body,
  its four subsection groups, and the cost table), replaced by a one-line
  pointer under the retained heading; gained the visible handbook line under
  its H1, and its intro sentence now sends the reader to the leaf for the
  invariants. Everything else in the file is untouched, since three sibling
  leaves are splitting the rest of it in parallel.
- **`README.md`** — the `## Documentation` list splits "series invariants" out
  of the `BRANCHES.md` entry into an entry of its own naming the leaf.
- **`AGENTS.md`** — the routing table does the same: the `BRANCHES.md` row
  keeps the branch model and flavors, a new row routes "what every series
  guarantees, and what enforces it" to the leaf.
- **`scripts/EACH.md`** — its two citations of `BRANCHES.md#ci--green` (C1)
  and `BRANCHES.md#source-of-truth-cross-series` (G1) now point at the leaf;
  those anchors ceased to exist with the section.

## Assumptions

- **"Nothing enforces this" is a claim I checked, not an inference.** For each
  invariant I searched `.github/workflows/`, `scripts/`, and `tests/` for a
  check. Worth a reviewer's confirmation on the four negatives that matter
  most: no script computes `git cherry main dev` in S4's direction
  (`series-port.sh` runs the opposite direction — what `main` has and the
  series lacks); nothing compares `stable` against `lts`; nothing asserts a
  version prefix; and nothing detects a merge commit.
- **`main` is the only protected branch in `duckdb/duckdb-r`** — read from the
  branches API. What that protection actually requires (linear history or not)
  is repository configuration I cannot see, so the page says only that L is
  not computed anywhere in the tree and leaves protection out of the claim.
- **C1's real scope is `{S}-green`, not `{S}-dev`.** The source states "every `dev`
  commit is green"; the machinery guarantees it up to `{S}-green`, with
  `{S}-green..{S}-dev` in flight and undecided and `{S}-build` not built at
  all. I wrote the narrower, true version and said why.
- **C2 is partial.** `main` is checked on push, PR and nightly; `v*` tags and
  `cran-*` branches are checked by `R-CMD-check-dev.yaml`; no workflow
  triggers on `v1.4-andium`, `v1.4-andium-lts` or `v1.5-variegata`, so between
  releases those tips are green by provenance only.
- **G2's mechanism has a limit I stated as a boundary.** `vendor-one.sh`
  deletes any patch that fails `patch --forward --dry-run`, which cannot
  distinguish a patch that landed upstream from one that merely stopped
  applying. That is also in tension with `BRANCHES.md` § Patch Stack, which
  says such a run should fail and the patch be updated; I described the code's
  behavior and flagged a dropped patch as something to check.
- **A2's "linearization runbook" does not exist as a document.** I looked for
  it and say so rather than guessing that `series-forward.md` / `series-rebase.md`
  is meant.
- **Point-in-time figures were kept but marked.** The ~170 grandfathered
  merges, the 402 pending commits, the stray merge on `v1.5-variegata-dev`,
  the `dev-base` version readings — none is re-verified here (the `-dev`
  branches live in `krlmlr/duckdb-r`), and the closing section says plainly
  that these are observations, not values a check keeps current.
- **`AGENTS.md` and `README.md` will conflict.** Three sibling leaves are
  splitting `BRANCHES.md` at the same time and each needs the same rows
  edited; resolving on merge is expected.
- **`scripts/EACH.md` will conflict harder than that.** #2485 deletes the file
  outright, having absorbed it into `operations/ci/per-commit/`. The two link
  repoints made here go away with it; nothing else in this branch depends on
  the file surviving.

No durability sweep was run on this branch: the leaf's numbers are the
C-numbered identifiers and the point-in-time observations named above, both of
which the sweep's rules keep.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,7 +12,8 @@ start there rather than searching.
 | To solve | Read |
 |---|---|
 | Build, test, and validate the package locally | this file (below) |
-| Branch model, package flavors, series invariants | `BRANCHES.md` |
+| Branch model, package flavors | `BRANCHES.md` |
+| What every series guarantees, and what enforces it | [`handbook/branches/invariants/`](/handbook/branches/invariants/README.md) |
 | Release process, modelled as a state machine | `RELEASE.md` |
 | Vendoring mechanics: scripts, invariants, troubleshooting | `scripts/VENDORING.md` |
 | Per-commit CI (sharded matrix): design and limits | `scripts/EACH.md` |
--- a/BRANCHES.md
+++ b/BRANCHES.md
@@ -1,10 +1,13 @@
 # Branching Strategy
 
+*Handbook: [`branches/invariants/`](/handbook/branches/invariants/README.md).*
+
 Version numbers are given at the time of writing (March 2026) and may be outdated by the time you read this.
 The branching strategy is expected to remain stable.
 
-This document defines the branch model, the package flavors, and the
-**[series invariants](#series-invariants)**. For the release process itself —
+This document defines the branch model and the package flavors; the
+**[series invariants](/handbook/branches/invariants/README.md)** are in the
+handbook. For the release process itself —
 modelled as a finite state machine — see [`RELEASE.md`](RELEASE.md).
 
 ## Package Components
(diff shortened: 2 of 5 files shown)


krlmlr:
`main` now carries a written leaf for the series invariants at core depth (#2493), grouping the statements into families and pointing at `BRANCHES.md` for the statements themselves. This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Salvaged from here: the per-invariant enforcement audit. The leaf says most invariants are enforced by nothing and defers the per-statement notes to `BRANCHES.md`, which does not carry them; this branch derived them against the scripts, and the shape held — the only continuous mechanical checks are `series-advance.sh`'s ancestor gate before it pushes `-green` or `-build-base`, and the vendor counter. Also salvaged as a defect: `BRANCHES.md` cites a linearization runbook that does not exist as a document in the repository.

---

---

PR #2479, opened 2026-08-02
docs(handbook): Write the build/configuration leaf

krlmlr:
## What this leaf now owns

Every knob that changes how the package builds, as a reference table: one row per knob with what it does, its default, and when to set it. Build-time knobs (`MAKEFLAGS`, `NOT_CRAN`, ccache, `UserNM`, `DUCKDB_R_PREBUILT_ARCHIVE`, `DUCKDB_R_LIB_DIR`, `PKG_BUILD_EXTRA_FLAGS`, `DUCKDB_R_POISON_ENGINE`) are separated from vendoring-time knobs read by `scripts/rconfigure.py` (`DUCKDB_R_EXTENSIONS`, `DUCKDB_DEBUG_MOVE`, `DUCKDB_R_LINENR`, `DUCKDB_PATH`), because setting the latter at install time does nothing. A closing section names the `DUCKDB_R_*` variables that are *not* build knobs and links each to the leaf that owns it. Boundaries stated: `DUCKDB_R_USE_SYSTEM_LIB` is named and linked to `build/fast-paths/` but not explained here; `UserNM`'s `R CMD check` hazard is cross-referenced to `build/source-build/`; what the CI matrix sets is left to `operations/ci/matrix/`; the no-warning-suppression policy to `architecture/glue/`.

Every default was read out of the code that supplies it — `configure`, `configure.win`, `src/Makevars.in`, the `.mk` includes, `scripts/setup-makeflags.R`, `scripts/rconfigure.py`, `tools:::read_symbols_from_object_file()`, `pkgbuild:::should_add_compiler_flags()`, and the `.github/workflows/` actions.

## `DUCKDB_BUILD_UNITY` and the `DUCKDB_R_BINDIR` triple are gone

An earlier revision of this leaf documented two further vendoring-time knobs, and this PR depended on #2477 for one of them. #2477 now **deletes** both from `scripts/rconfigure.py` instead:

* `DUCKDB_BUILD_UNITY` reached upstream's `build_package()` as `unity_count`, and `generate_unity_builds()` has accepted that argument without reading it since well before v1.0.0 — unity grouping is decided per directory by `add_library_unity`. A knob that is read, validated and then disregarded is worse than no knob.
* `DUCKDB_R_BINDIR`/`DUCKDB_R_CFLAGS`/`DUCKDB_R_LIBS` wrote a `src/Makevars` for a system-library build that has not worked since `src/Makevars.in` lost its `{{ SOURCES }}` placeholder.

Both rows and the paragraph explaining them are therefore removed here. A knob table lists what exists; the removals are recorded in #2477, not in this leaf. Nothing in this PR now depends on the ordering with #2477 — the two are consistent whichever merges first, except that this leaf briefly stops naming variables that `main` still reads.

## What moved

* `AGENTS.md` — lost the `export MAKEFLAGS="-j$(nproc)"` bullet; gained a one-line pointer to the leaf in its place and the visible italic backreference under its H1. Nothing else was touched, since five sibling leaves are absorbing the rest in parallel.
* `README.md` — lost the two knob sentences under "Building" (`MAKEFLAGS` for parallel builds, configure `ccache`); gained a pointer to the leaf in their place.
* `scripts/README.md` — unchanged: `.claude/skills/docs-consistency/docs-readme.R` already maps `setup-makeflags.R` to `handbook/build/configuration`, so the generated index already carries the backreference.

## Assumptions

* The stub named four knobs; I read "that list is a starting point" as licence to sweep `configure`, `configure.win`, `src/Makevars.in`, the `.mk` includes, `scripts/rconfigure.py`, `scripts/setup-makeflags.R` and `.github/workflows/` for every variable the build reads, and to document all of them. The two tables name what came out.
* `PKG_BUILD_EXTRA_FLAGS` is pkgbuild's variable, not this package's, but CI sets it and it changes the compiler flags a `load_all()` build uses — so I judged it a build knob. Confirm that it belongs here rather than in `contributors/setup/`.
* `DUCKDB_R_LIB_DIR` is only read while `DUCKDB_R_USE_SYSTEM_LIB` is active. I kept it here (it is a variable `configure` reads) and linked the mode itself to `build/fast-paths/`; a reviewer may prefer it live entirely in that leaf.
* `DUCKDB_R_POISON_ENGINE` is set by a matrix entry and consumed by a CI action, not by `configure`. I gave it a row because its effect is a compile-time macro, and pointed both at `operations/ci/matrix/` for who sets it and `testing/guards/` for what it proves.
* `UserNM=true`'s saving was asserted by `AGENTS.md` and the comment in `.github/workflows/install/action.yml`, not measured here; the leaf no longer quotes it — see the durability sweep below.
* `DUCKDB_R_LINENR`'s effect is decided by upstream's `package_build.py`, which is not in this repository. The row says so instead of guessing.
* Trimming the root `README.md` was not in the stub's work order, but those two sentences are knob prose and "move, don't copy" applies. Flagging it in case the root `README.md` is meant to stay self-contained for CRAN readers, where `/handbook/…` does not resolve.

## Durability sweep

**Out.** The "~10–20 s" for `UserNM=true`, which was a figure quoted from
`AGENTS.md` and a CI comment rather than measured; the row now says the knob
trims a step from the end of the install, which is what a reader needs to
decide. Three counts of knobs: "the two statically linked in every build"
(both are named on the same line), "a fourth vendoring-time knob" (it counted
the table above it, which has since grown), and "Three more are outputs rather
than inputs" (the list follows immediately). This PR body's own count of the
knobs went the same way; the tables name each one, which is the durable form.

**Stayed.** Every knob name, default and file path in the tables — the tables
are the mechanism a reader consults, and they age with the tree rather than
against it. The two-core `MAKEFLAGS` cap, which is CRAN policy and the reason
`NOT_CRAN` exists. `R >= 4.2` for the Windows archive-reuse branch and
`CXX_STD = CXX17`, which identify a version rather than counting one. ccache
"from minutes into seconds" is qualitative and survives drift.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,7 @@
 # DuckDB R Package - Operational Instructions
 
+*Handbook: [`build/configuration/`](/handbook/build/configuration/README.md).*
+
 R package that contains a vendored copy of the DuckDB C++ library and glue code for R, including a DBI and a relational interface.
 
 ## Where to look
@@ -25,7 +27,7 @@ start there rather than searching.
 
 - `sudo apt-get install -y r-base r-base-dev build-essential` -- installs R and development tools
 - `mkdir -p ~/R/library && echo '.libPaths("~/R/library")' > ~/.Rprofile` -- sets up local R library
-- `export MAKEFLAGS="-j$(nproc)"` -- enables parallel compilation
+- Build knobs -- `MAKEFLAGS` and parallelism, ccache, `UserNM`, the object-file archive, and everything else the build reads from the environment: [`handbook/build/configuration/`](/handbook/build/configuration/README.md)
 - `UserNM=true R CMD INSTALL . --no-byte-compile` -- builds and installs the package. NEVER CANCEL: takes 10-15 minutes on first build. Set timeout to 30+ minutes.
 
 `UserNM=true` is an install-time shortcut only -- **never export it for `R CMD check`**.
--- a/README.md
+++ b/README.md
@@ -114,8 +114,9 @@ Then, install:
 ~duckdb-r: R CMD INSTALL .
 ```
 
-Set the `MAKEFLAGS` environment variable to `-j8` or similar for parallel builds.
-Configure `ccache` for faster repeated builds.
+The build knobs — parallelism, `ccache`, the object-file archive, and the
+rest — are documented in
+[`handbook/build/configuration/`](/handbook/build/configuration/README.md).
 
 If you wish to test new DuckDB functionality with duckdb-r, make sure your clone of `duckdb-r` is one level deeper than your clone of `duckdb` (e.g. `R/duckdb-r` and `duckdb`).
 Then run the following commands:
--- a/handbook/build/configuration/README.md
+++ b/handbook/build/configuration/README.md
@@ -1,19 +1,123 @@
 # Configuration
 
-*Stub — this leaf will own its topic;
-today it routes to where the knowledge lives.
-The writing protocol is in [`meta/handbook/`](/handbook/meta/handbook/);
-the last section holds this leaf's parameters.*
+The knobs that change how the package builds:
+the environment variables `configure` and `src/Makevars` read,
+the settings that live in `~/.R/Makevars` rather than in the environment,
+and the variables the vendoring script reads when it *generates*
+`src/Makevars`.
+What the build then does with them is
+[`build/source-build/`](/handbook/build/source-build/README.md)'s topic;
+the runtime settings that share the `DUCKDB_R_` prefix are not build knobs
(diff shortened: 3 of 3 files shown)


krlmlr:
`main` now carries a written `build/configuration/` leaf (#2493), splitting the knobs into those read at build time and those read at vendoring time, and saying why that split matters. This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Nothing was salvaged from here. Its one candidate was the `UserNM` mechanism — that R's symbol reader takes `nm` from that variable and `tools:::check_so_symbols()` uses the same reader — but the leaf already states the consequence that matters, that it blinds the check's symbol scan and must never be exported for `R CMD check`.

---

---

Issue #1052, opened 2025-02-13
Remove `allow_materialization` argument


krlmlr:
No longer needed, but double-check.

Issue #98, opened 2024-03-04
Make `duckdb_register` as a generic function


eitsupi:
Currently there is a separate function named `duckdb_register_arrow` available apart from `duckdb_register`, but I am wondering if these can be made into one S3 generic function.

krlmlr:
Thanks. What other use cases do you have in mind?

eitsupi:
> What other use cases do you have in mind?

Not that I have any other use cases, but I felt that I didn't want to use another function when supporting Apache Arrow implementations outside of the `arrow` package in the future.

krlmlr:
What do you mean by "supporting Apache Arrow implementations" -- nanoarrow?

What other table-like objects would we want to register?

eitsupi:
I assumed something like nanoarrow array or polars DataFrame.
Of course to support them, the C++ code in the duckdb package would have to be changed, I suppose.

krlmlr:
But if we change the C++ code anyway, we can surely add a new `duckdb_register_*()` function?

krlmlr:
Working on a proof of concept what scanning of other data sources could look like.

PR #2480, opened 2026-08-02
docs(handbook): Write the architecture/r-layer leaf

krlmlr:
## What this leaf now owns

[`handbook/architecture/r-layer/`](/handbook/architecture/r-layer/README.md)
explains the R side of the package:
the one-file-per-method layout under `R/`
(the `{generic}__{signature}.R` files, no `setMethod()` anywhere else),
the generated-and-committed files and their regenerators
(`R/cpp11.R` from `cpp11::cpp_register()`,
`R/rethrow-gen.R` from `scripts/rethrow.R` via `.Rprofile`,
`R/version.R` from `scripts/rconfigure.py`),
the `get_package_name()` flavor seam in `R/package.R`
and why nothing may hard-code the package name,
the S4 and S3 classes,
and deferred S3 registration for suggested packages in `.onLoad()`.
Boundaries state today's behavior for
[#98](https://github.com/duckdb/duckdb-r/issues/98) and
[#1052](https://github.com/duckdb/duckdb-r/issues/1052);
neither is closed.
The C++ side, the flavor *model*, and the flavor-name guard as a test
are linked to `architecture/glue/`, `branches/flavors/`, and
`testing/guards/` rather than repeated.

(The file-name pattern is written with braces here. The leaf spells it
`{generic}__{signature}.R` with angle brackets; GitHub deletes those from a
pull-request body even inside backticks.)

## What moved

* `AGENTS.md` — lost the body of § "Never Hard-Code the Package Name"
  (absorbed into the leaf); the heading stays as a one-line pointer.
  Gained the visible italic handbook backreference under its H1.
  Not deleted: it is the entry point coding agents load by convention.
* `R/package.R` — gained a plain-comment backreference to the leaf,
  above the `get_package_name()` seam.
* `R/zzz.R` — gained a plain-comment backreference to the leaf,
  above `.onLoad()`.
* `scripts/rethrow.R` — unchanged. The generated `scripts/README.md`
  already groups it under `architecture/r-layer/`, so it needs no
  comment of its own and no regeneration was required.

## Assumptions

* **`AGENTS.md` is being split by several leaves in parallel**, so only
  the one section was taken. The heading was kept with a pointer rather
  than removed outright, per the partial-absorption rule.
* **The root `README.md` was left alone.** Its `## Documentation`
  section lists `AGENTS.md`, which is not going away; updating that
  entry seemed to belong to whoever finishes the split, not to a
  partial absorber.
* **Backreference comments went into two files, not all of `R/`.**
  `R/package.R` and `R/zzz.R` are the files this leaf is specifically
  about. `R/cpp11.R` and `R/rethrow-gen.R` were deliberately left
  untouched — they are generated and say so, and a comment there would
  be wiped by the next regeneration.
* **The file-per-method convention is described as covering S4 methods
  only**, because the directory contradicts a stronger reading: the
  adbcdrivermanager S3 methods live in `R/Driver.R`, the dplyr/dbplyr
  ones in `R/backend-dbplyr__duckdb_connection.R`, and the relational
  S3 methods in `R/relational.R`. Two `__` files (`print__duckdb_explain.R`,
  `backend-dbplyr__duckdb_connection.R`) carry S3, so the page reads the
  `__` name as "methods for one class" rather than "S4".
* **#98's verdict is read from the issue thread** — the maintainer's
  closing comment ("if we change the C++ code anyway, we can surely add
  a new `duckdb_register_*()` function?"). The leaf states that as the
  reasoning against a generic while leaving the issue open. Worth
  confirming that is still the position.
* **#1052 was treated as a pending code fix.** `rel_to_altrep()` in
  `R/relational.R` still accepts `allow_materialization` and folds
  `FALSE` into `n_cells = 0`; the leaf documents that as today's
  behavior. The claim that duckplyr is the caller to coordinate with
  comes from the issue's `duckplyr 🗜️` label and the `@noRd` /
  "FIXME: Move dots after `rel` for duckplyr >= 1.1.0" comment in the
  source, not from reading duckplyr.
* **Nothing was built.** All claims come from reading `R/`, `.Rprofile`,
  `scripts/rethrow.R`, `scripts/flavor.sh`, `scripts/flavor.patch`,
  `NAMESPACE`, and `DESCRIPTION`. The roxygen-inline-chunk failure mode
  and the `R/storage.R` anecdote are carried over from `AGENTS.md` as
  asserted there.

## Durability sweep

This leaf held the most of them, because most of its content is inventory.

**Out:**

- *"Forty-four files register S4 methods this way"* → the mechanism: the
  `{generic}__{signature}.R` naming convention *is* the index, and there
  is no `setMethod()` call elsewhere in `R/`. Adding one method used to
  falsify the sentence.
- *"Three files under `R/` are generated"* → *"A few"*; the table below
  it names each one, which is the durable form.
- *"Five S4 classes, all exported"* → *"The S4 classes, all exported"*;
  the table names them.
- Bare counts in front of lists that already name their members:
  *"the three adbcdrivermanager methods"*, *"registers four"* signatures
  of `dbQuoteIdentifier`, *"Their seven methods"*, *"those ten methods"*,
  *"six dbplyr generics"*, *"three adbcdrivermanager generics"*, and
  *"the four rlang functions"*. The deferred-registration paragraph now
  points at the `s3_register()` calls in `.onLoad()` as the list.
- *"exposes the first two as `$pkg` and `$env`"* → the helper names, so
  the sentence no longer depends on the order of the table above it.

**Kept:**

- The **named lists themselves** — the four S3 classes, the three names
  doing double duty as class and constructor, the two `__`-named S3
  files, the four rlang fallbacks. Each names its members inline, so a
  change to the set forces a change to the prose; that is a list, not a
  hostage.
- *"two plain functions, not one generic"* — here the count *is* the
  design question the boundary section is about.
- *"The hierarchy is DBI's, one level deep"* and the hard dependency set
  (`DBI`, `methods`, `utils`, and nothing else), both named and
  load-bearing.

---


krlmlr:
`main` now carries a written leaf for the R layer at core depth (#2493), so this branch predates the tree it edits, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Salvaged from here: the roxygen failure mode. The leaf states the rule — inline chunks go through the package-name seam, never a `:::` qualifier — but not what happens when one slips through. roxygen2 evaluates inline chunks in the package's own namespace and an unresolvable reference does not raise: the chunk *and every other inline chunk in the same block* are emitted verbatim as `\verb{r ...}`, so one qualified reference in `R/storage.R` also took out the `lifecycle::badge()` chunk beside it, and CI fails at `roxygenize` before anything is compiled.

---

---

Issue #2234, opened 2026-03-17
Toolchain update for Windows builds of DuckDB


staticlibs:
Hi R client maintainers!

We are changing the toolchain that is used on DuckDB CI to produce Windows binaries, details:

 - duckdb/duckdb#21371
 - duckdb/extension-ci-tools#343

Currently the following MinGW build is used:

```
https://github.com/niXman/mingw-builds-binaries/releases/download/14.2.0-rt_v12-rev0/x86_64-14.2.0-release-posix-seh-ucrt-rt_v12-rev0.7z
```

The the GCC version `14.2` was chosen to stay closer to release Linux builds.

We can change the toolchain if necessary, please let us know if it is better to use different kind/version of MinGW binaries. Besides R these binaries are also used by Golang client.

abfleishman:
Ooo would this help enable some of the extensions that don't work on windows listed in #100!?

Upipa:
If I understood well this would enable the motherduck extension. I have a blocked project for this extension missing :(

abfleishman:
Any updates here?

krlmlr:
@staticlibs: For a related issue (extension loading on Linux), I have discovered that compiling duckdb with `-stdlib=libc++` instead of `-stdlib=libstdc++` breaks extensions: https://github.com/duckdb/duckdb-r/issues/1107#issuecomment-5059454255 . Does this ring a bell?

This is nothing the package can control, just like the Windows toolchain. I wonder what the appetite is in general to support multiple flavors per OS. 

I have also investigated the possibility of just using the C interface of DuckDB: feature is currently blocked by the missing "relational" API and some Arrow pushdown details. Is the expansion of the C API on the roadmap? Who could support here?

krlmlr:
Also: is there a robust way to obtain an extension's source for a particular version of DuckDB, to compile locally?

staticlibs:
@krlmlr 

Hi! 

> Also: is there a robust way to obtain an extension's source for a particular version of DuckDB, to compile locally?

This is a deep topic, in general, if you are building multiple extensions for a particular version of DuckDB, the easiest way to do so is something like this from `duckdb/duckdb` working copy:

```bash
[...]

Issue #1581, opened 2025-09-18
INSTALL spatial fails on windows


fvalenduc:
Can somebody explain this ?

> dbExecute(con,"INSTALL spatial")
Error in `duckdb_result()`:
! Invalid Error: HTTP Error: Failed to download extension "spatial" at URL "http://extensions.duckdb.org/f40ac6e565/windows_amd64_mingw/spatial.duckdb_extension.gz" (HTTP 403)
Extension "spatial" is an existing extension.

For more info, visit https://duckdb.org/docs/stable/extensions/troubleshooting/?version=f40ac6e565&platform=windows_amd64_mingw&extension=spatial
ℹ Context: rapi_execute
ℹ Error type: INVALID
ℹ Raw message: HTTP Error: Failed to download extension "spatial" at URL "http://extensions.duckdb.org/f40ac6e565/windows_amd64_mingw/spatial.duckdb_extension.gz" (HTTP 403)
Extension "spatial" is an existing extension.

For more info, visit https://duckdb.org/docs/stable/extensions/troubleshooting/?version=f40ac6e565&platform=windows_amd64_mingw&extension=spatial
Run `rlang::last_trace()` to see where the error occurred.

This is on a Windows VM running on a Linux host. On the linux host, it works perfectly fine.

fvalenduc:
> dbExecute(con,'INSTALL SPATIAL;')
Error in `duckdb_result()`:
! Invalid Error: HTTP Error: Failed to download extension "spatial" at URL "http://extensions.duckdb.org/v1.4.0/windows_amd64_mingw/spatial.duckdb_extension.gz" (HTTP 403)
Extension "spatial" is an existing extension.

For more info, visit https://duckdb.org/docs/stable/extensions/troubleshooting?version=v1.4.0&platform=windows_amd64_mingw&extension=spatial
ℹ Context: rapi_execute
ℹ Error type: INVALID
ℹ Raw message: HTTP Error: Failed to download extension "spatial" at URL "http://extensions.duckdb.org/v1.4.0/windows_amd64_mingw/spatial.duckdb_extension.gz" (HTTP 403)
Extension "spatial" is an existing extension.

For more info, visit https://duckdb.org/docs/stable/extensions/troubleshooting?version=v1.4.0&platform=windows_amd64_mingw&extension=spatial
Run `rlang::last_trace()` to see where the error occurred.

krlmlr:
[...]

Issue #100, opened 2024-03-05
Multiple extensions missing for Windows (azure, iceberg, spatial, +)


thohan88:
Hi,
Thanks for all the great work on maintaining this package. I can interpret from the work being done on 0.10.0 that this requires from effort, but it's greatly appreciated. 

I have noticed that there are multiple extensions missing on Windows as reported in other issues (#23). I made a quick script to check which extensions are missing based on `duckdb_extensions()` (0.9.2) and came up with the following results for version 0.10.0 and 0.9.2:
# 0.10.0
<table>
  <thead>
    <tr>
      <th rowspan="2" colspan="1">Extension</th>
      <th rowspan="1" colspan="3">
        <span>Linux</span>
      </th>
      <th rowspan="1" colspan="2">
        <span>OSX</span>
      </th>
      <th rowspan="1" colspan="2">
        <span>Windows</span>
      </th>
    </tr>
    <tr>
      <th rowspan="1" colspan="1">amd64</th>
      <th rowspan="1" colspan="1">amd64_gcc4</th>
      <th rowspan="1" colspan="1">arm64</th>
      <th rowspan="1" colspan="1">amd64</th>
      <th rowspan="1" colspan="1">arm64</th>
      <th rowspan="1" colspan="1">amd64</th>
      <th rowspan="1" colspan="1">amd64_rtools</th>
    </tr>
  </thead>
  <tbody>
    <tr><td align = "middle">arrow</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">❌</td>
<td align = "middle">❌</td></tr>
    <tr><td align = "middle">autocomplete</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td></tr>
    <tr><td align = "middle">aws</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">❌</td></tr>
    <tr><td align = "middle">azure</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
<td align = "middle">✅</td>
[...]

PR #2481, opened 2026-08-02
docs(handbook): Write the source build leaf

krlmlr:
## What this leaf now owns

`handbook/build/source-build/` explains the real build end to end: the bootstrap requirements, the two stages (`configure`, then `make` in `src/`), the three `Makevars.*` fragments `configure` writes and the order it writes them in, how `src/Makevars` and `src/Makevars.win` are generated from `src/Makevars.in` by `rconfigure.py` and why their include order is load-bearing, the `%.dd: %.d` dependency rule and when a `.dd` change is spurious, the `cleanup` → `src/duckdb.tar.xz` → `configure` tarball round trip together with what `.Rbuildignore` withholds, and the `UserNM` shortcut with the reason it must never be exported for `R CMD check` (R reads `nm` from `$UserNM` in `tools:::read_symbols_from_object_file()`, which both the `symbols.rds` rule and `tools:::check_so_symbols()` go through, so `true` blinds the registration scan). A Windows section covers `configure.win`, the R 4.2 `librstrtmgr` split, and answers [#2234](https://github.com/duckdb/duckdb-r/issues/2234): the toolchain is whatever Rtools the installing R ships, so an upstream MinGW change does not alter how the package compiles — it alters which prebuilt artifacts match the `windows_amd64_mingw` platform the Rtools build reports. The fast path, the individual knobs, what generates `src/duckdb/`, and CRAN's tarball policy are linked, not repeated.

## What moved

* `AGENTS.md` — lost the body of "Bootstrap, Build, and Test the Repository" (including the `UserNM` warning), the `### Building` command block under "Common Tasks", and the `.dd` paragraph; each is now a one-line pointer to the leaf. Gained the visible italic handbook backreference under its H1. Not deleted: it is the entry point coding agents load, and five other leaves are splitting the rest of it in parallel.
* `configure`, `configure.win`, `cleanup`, `cleanup.win` — gained a `# Handbook: handbook/build/source-build/` comment below the shebang.
* `src/Makevars.in` — gained the same comment below its `{{ HEADER }}` line.
* `src/Makevars`, `src/Makevars.win` — gained the identical line, so the checked-in generated files stay byte-identical to what `rconfigure.py` will emit from the amended template on the next vendor run.
* `src/include/deps.mk`, `from-tar.mk`, `to-tar.mk`, `to-tar-win.mk` — gained the same comment as the first line.
* **`R CMD build` empties the tree it runs in** — the caution arrived here from the vendoring pipeline leaf (#2477), where it was not the topic. Restated in this leaf's voice, beside the `cleanup` script it is about: `cleanup` runs in place, `git clean -fdx src` takes the untracked and ignored build output with it, and `src/duckdb/` is archived and removed, so a tree in the middle of an edit-build-test loop or a vendor replay is exactly the tree not to build a tarball from.

## Assumptions

* I took only my section of `AGENTS.md`, plus two paragraphs that are unambiguously this leaf's topic but sit elsewhere in the file: the `### Building` code block under "Common Tasks" (leaving a bare `UserNM=true` invocation there without its warning would be worse than a pointer) and the `.dd` paragraph, which sits inside the fast-path section that another leaf is absorbing in parallel. Expect a merge conflict there; both resolutions are fine.
* `AGENTS.md`'s "Where to look" table still routes "Build, test, and validate the package locally" to "this file (below)", which now routes on to the leaf. I left the table for whoever rewrites it wholesale rather than editing one row of a shared table from five branches at once.
* The root `README.md` was not touched. Its `## Building` and `## Documentation` sections are user-facing surfaces served by several leaves, and `AGENTS.md` — still listed there — now carries the handbook backreference itself.
* Hand-editing `src/Makevars` and `src/Makevars.win`, which say "do not edit by hand", was deliberate: the inserted line is exactly what the amended `src/Makevars.in` produces, so the three files stay consistent until the next regeneration. Worth a reviewer's confirmation.
* Nothing was built. Every claim is read from `configure`, `configure.win`, `src/Makevars.in`, the `.mk` fragments, `cleanup`, `.Rbuildignore`, `scripts/rconfigure.py`, `scripts/setup-makeflags.R`, R's own `share/make/shlib.mk`, and the deparsed `tools:::read_symbols_from_object_file()` / `tools:::check_so_symbols()`. Two readings are mine rather than the sources': that `tar --force-local` on Windows is there so a drive letter is not taken for a remote host, and that `Makevars.rstrtmgr` having no fallback rule is why make stops early without `configure` (the second is corroborated by a comment in `scripts/vendor-one.sh`).
* #2234 stays open and uncommented; the leaf states today's behavior and links it as a boundary. Its two sub-issues (#100, #1581) are extension-availability questions and were left to `usage/extensions/`.
* `AGENTS.md`'s `^ *` for the fixed `deps.mk` filter did not match the file, which uses `^ +`; the leaf states what `src/include/deps.mk` actually contains.

## Durability sweep

A follow-up pass removed the numbers a routine commit invalidates.

**Out:**

* The first build's "10–15 minutes, budget 30 minutes" → "tens of
  minutes, set the timeout generously". The instruction is not to cancel
  it, and that survives a faster compiler.
* "(fifteen today)" after `include/glue.mk` → gone. The sentence already
  names the file that lists the glue objects, which is the durable fact.
* "saves some 10–20 seconds" → "saves a few seconds"; the point is that
  the saving does not justify the `R CMD check` hazard.

**Kept**, because each is an enumeration the design rests on and is
spelled out in the same breath: the three `Makevars.*` fragments,
`configure.win`'s three differences, `src/Makevars.in`'s three
placeholders, the two halves of the object-archive cache, the three ways
`cleanup` is careful. Also kept CRAN's 5000000-byte tarball limit — a
policy threshold the `cleanup` scripts cite, not a measurement of this
tree — and the `^  ` / `^ +` filter patterns, which are code.

---


krlmlr:
`main` now carries a written `build/source-build/` leaf (#2493), with the two stages, what `configure` writes, why it deletes a stale `Makevars.system-lib` first, and the `.dd` files. This branch predates that, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Nothing was salvaged from here directly. Its most useful observation — that `src/include/deps.mk` records only local headers — survives, but through #2477, where it is stated with the consequence that makes it actionable: a local commit-by-commit vendor replay that does not delete `src/*.o` links glue built against a different engine.

---

---

PR #2483, opened 2026-08-02
docs(handbook): Record the steering taken during the first leaf wave

krlmlr:
**Standing PR for handbook steering.** Decisions taken while the first
wave of 26 leaves is being written land here, so the rules stay ahead of
the leaves instead of being reconstructed afterwards. Expect this branch
to grow; it is the one place to look for "what did we decide, and why".

## Decisions in this PR

**Leaves are comprehensive; pointing is the exception.** The previous
wording — "Pointer leaves are legitimate" — invited a leaf to defer to a
reference page, and one did: `usage/storage/` came back as a thin
pointer to `?duckdb_storage`. That contradicts the rule two bullets
down, which already calls reference pages *secondary*. Now stated
directly: a reference page is derived from its leaf, not the leaf's
excuse to be thin. Only two homes are genuinely elsewhere —
documentation that is not ours (upstream DuckDB docs) and the
machine-loaded playbooks under `.claude/skills/`, whose procedures are
executed rather than read.

**Intent lives outside the tree.** Plans stay in `plan/`, deliberately.
`meta/plans/` explains that directory and its conventions; each affected
leaf links the plan carrying its intent. And a plan that has become fact
is documented as fact in the leaf, with no trace of its having once been
a proposal.

**Two additions to the backreference forms:**

* *Generated files* — the backreference belongs in the **generator**, in
  the template the file is rendered from, so it survives the next
  regeneration. Editing the generated file is writing in sand. This came
  out of #2466, which put the line in `src/Makevars.in` and in the
  `R/version.R` template inside `scripts/rconfigure.py`.
* *Outbound correspondence takes none.* `cran-comments.md` is not
  documentation of a topic but a letter whose whole text reaches a CRAN
  maintainer verbatim. The link to it is one-way, and that asymmetry is
  correct rather than an orphan. Came out of #2474, which added one and
  then reverted it.

## Already landed separately

* #2458 — no tombstones (an absorbed file goes away; the in-place
  generated `README.md` routes on), and every upward link written from
  the repository root.

## Open, not decided here

* **How `AGENTS.md` carries its backreference.** Agents split on this:
  most added the italic H1 line, #2473 argued the file's own
  `## Where to look` routing table *is* its backreference mechanism and
  updated rows instead. Six leaves touch that file in this wave, so
  whichever lands first sets the precedent. Worth settling explicitly
  rather than by merge order.
* **Extending the generated index beyond `scripts/`.** #2459 generalises
  `docs-readme.R` to cover `plan/` as well; the remaining directories
  that hold documentation are still open.

---

### DIFF
--- a/handbook/meta/handbook/README.md
+++ b/handbook/meta/handbook/README.md
@@ -213,6 +213,23 @@ because nothing in the sentence breaks when the forty-fifth lands.
 A list is safe, a tally is not,
 and the same number can be either depending on what surrounds it.
 
+**A fact known to be provisional carries the link to what will change it.**
+The rules already say a leaf points at the plan
+that carries its intent;
+intent also lives as an issue, ours or upstream,
+and those count the same.
+Without the link a reader cannot tell
+"this is how it works" from "this is how it works for now",
+and the two call for different decisions.
+With it, the leaf records that we already know —
+so the next reader does not re-derive the finding,
+and the acknowledgement is not lost
+the way it is when it lives only in a code comment
+or a pull-request thread.
+A behaviour that survives only because nobody has fixed it yet
+is documented as such, with its issue,
+and stops being documented that way when the issue closes.
+
 A **default is a fact**, and belongs in the prose with its value.
 How many commits a pass consumes,
 how large a chunk the planner cuts,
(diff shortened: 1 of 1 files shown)


krlmlr:
`main` now carries `meta/handbook/` for the tree's structure and growth moves, and #2504 split the prose rules out into `meta/authoring/`, which is binding on every page. This branch predates that split, conflicts throughout, and merging it would revert the newer pair. Closing unmerged.

Nothing was salvaged from here. It is a rules page rather than a page about the code, so it carried no claim to verify, and the rule it proposed — never refer by position, name the thing — is now stated in `meta/authoring/` with the same reasoning.

---

---

Issue #1796, opened 2025-12-01
Calling `rapi_error_with_context()` is unsafe from ALTREP methods


krlmlr:
`rapi_error_with_context()` jumps into R, this appears to be unsafe or unsupported for ALTREP methods. I have seen stack overflows occur from this.

Define a guard that is activated whenever an ALTREP method is entered. Query this guard in `rapi_error_with_context()` and fall back to throwing an exception instead of calling into R via `rapi_error()` .

Issue #1147, opened 2025-05-06
R crashes if calling duck-ui with an overwritten connection object


mns-nordicals:
First create a duckdb connection and open the duck-ui

con <- DBI::dbConnect(duckdb::duckdb())
DBI::dbExecute(con, "CALL start_ui()")

close the browser and overwrite / create a new connection to duckdb

con <- DBI::dbConnect(duckdb::duckdb())

Now call the duckui again

DBI::dbExecute(con, "CALL start_ui()")

In positron I can get this error message / log

2025-05-06 23:37:53.749 [info] ** Begin kernel log for session R 4.5.0 (r-efa4682b) at 5/6/2025, 11:37:53 PM **
Streaming kernel log file: C:\Users\XXX\AppData\Local\Temp\kernel-G7luwG\kernel.log
terminate called after throwing an instance of 'duckdb::Exception'

  what():  {"exception_type":"Settings","exception_message":"Setting \"ui_polling_interval\" not found"}

terminate called after throwing an instance of 'duckdb::Exception'

This will also crash in Rstudio. 

I don't know if this is a duckdb-r issue, a duckdb issue or simply a user issue 😅

krlmlr:
Thanks. Can't reproduce with duckdb 1.3.2, R 4.4.3, latest RStudio on macOS. Is this still a problem?

krlmlr:
A `gc()` was needed to reproduce. This seems to be a duckdb-ui problem, please open an issue at https://github.com/duckdb/duckdb-ui. Workarounds:

- Explicitly shut down the UI
- Wait for a period larger than the `ui_polling_interval` setting (0.284 seconds)

``` r
## duckdb-r#1147 -- R crashes calling start_ui() with an overwritten
## connection object. The failure is an abort, not an R error, so each
## variant runs in its own process and we look at the exit status.
library(duckdb)
#> Loading required package: DBI
packageVersion("duckdb")
#> [1] '1.5.5.9012'

variant <- function(...) {
  script <- tempfile(fileext = ".R")
  writeLines(c("library(duckdb)", ...), script)
  out <- processx::run(
    file.path(R.home("bin"), "Rscript"),
    script,
    error_on_status = FALSE,
    stderr_to_stdout = TRUE
  )
  cat(out$stdout)
  cat("exit status:", out$status, "\n")
}

# 1. The reporter's sequence: the second connection lands on the same name,
#    so the first database is finalised while its UI server is still up
variant(
[...]

Issue #540, opened 2024-10-31
Rethink `rapi_rel_sql()`


krlmlr:
https://github.com/duckdb/duckdb/pull/14629#issuecomment-2450301365

Is our implementation doing too much? The code example is this:

```r
con <- DBI::dbConnect(duckdb())
rel <- rel_from_df(con, mtcars)
rel2 <- rel_sql(rel, "SELECT hp, cyl FROM _ WHERE hp > 100")
```

The relation is converted to a view named `_`, and then a query is run on this and materialized as an R data frame.

To be honest, I don't think this is used a lot, and it's not exported. I don't have access to https://github.com/duckdblabs/duckdb-internal/issues/1721 (referred to from https://github.com/duckdb/duckdb-r/pull/138), so I may not understand the importance of this.

We definitely should make it easy to create a table (and perhaps also a view) from a relational object, and also to represent a table or view as a relational object. When this is done, this function can be deleted I think.

Issue #101, opened 2024-03-10
Construction of deep relational trees


krlmlr:
This is a toy example, but relevant for some CRAN packages with the default setting of `max_expression_depth` . The symptoms are the same as when evaluating `rel7` .

Ideally, we would already see an error when constructing `rel5` . However, the system lets me construct `rel5` and even `rel6`, only construction of `rel7` fails with the same error as the evaluation of `rel5` . Is this an off-by-two error, or something more serious?

duckplyr can fall back to dplyr if the error happens at construction, but not at evaluation -- this is too late. An error on construction of `rel5` or perhaps even `rel4` would fix the downstream problem. How to achieve this?

``` r
duckdb <- asNamespace("duckdb")
con <- DBI::dbConnect(duckdb::duckdb())
experimental <- FALSE
df1 <- tibble::tibble(id = 1L)

DBI::dbExecute(con, "SET max_expression_depth TO 5")
#> [1] 0


rel1 <- duckdb$rel_from_df(con, df1, experimental = experimental)
rel2 <- duckdb$rel_project(
  rel1,
  list({
    tmp_expr <- duckdb$expr_reference("id")
    duckdb$expr_set_alias(tmp_expr, "id")
    tmp_expr
  })
)
rel3 <- duckdb$rel_project(
  rel2,
  list({
    tmp_expr <- duckdb$expr_reference("id")
    duckdb$expr_set_alias(tmp_expr, "id")
    tmp_expr
  })
)
rel4 <- duckdb$rel_project(
  rel3,
  list({
    tmp_expr <- duckdb$expr_reference("id")
    duckdb$expr_set_alias(tmp_expr, "id")
    tmp_expr
  })
)
rel4
#> DuckDB Relation: 
#> ---------------------
#> --- Relation Tree ---
#> ---------------------
#> Projection [id as id]
#>   Projection [id as id]
#>     Projection [id as id]
#>       r_dataframe_scan(0x11cca4278)
#> 
#> ---------------------
#> -- Result Columns  --
#> ---------------------
#> - id (INTEGER)
rel5 <- duckdb$rel_project(
  rel4,
  list({
    tmp_expr <- duckdb$expr_reference("id")
    duckdb$expr_set_alias(tmp_expr, "id")
    tmp_expr
  })
)
rel5
#> DuckDB Relation: 
#> ---------------------
#> --- Relation Tree ---
#> ---------------------
#> Projection [id as id]
#>   Projection [id as id]
[...]

PR #2484, opened 2026-08-02
docs(handbook): Write the C++ glue leaf

krlmlr:
## What this leaf now owns

[`handbook/architecture/glue/`](https://github.com/duckdb/duckdb-r/blob/claude/handbook-arch-glue-zayxqn/handbook/architecture/glue/README.md)
is now written.
It owns the R ↔ DuckDB seam in `src/`: the glue translation units named as `GLUE` in `src/include/glue.mk` and what each is responsible for; `src/include/rapi.hpp` and the external-pointer wrappers (`DualWrapper`&#39;s precious/disposable handles, the garbage-collection warnings); the cpp11 entry points, the vendored cpp11 headers under `inst/include/cpp11/` and the `cpp11::cpp_register()` regeneration step; `RStrings` and *why* the rule about it exists (one preserved allocation instead of a `Rf_mkString()` per row), plus the distinction from `RStringsType`/`r_string`; the ALTREP relations — lazy materialization, the row/cell budget and its `LimitRelation` guard, the `duckdb.materialize_*` options, deferred error reporting, and the reverse direction through `AltrepDataFrameRelation`; the two paths an error takes back into R (`BEGIN_CPP11` frames and `rapi_error_with_context()`); and the C++ source conventions — what `scripts/format.py` formats and how, and the no-warning-suppression policy with its `patch/` remedy.

Issues drained: [#540](https://github.com/duckdb/duckdb-r/issues/540) (`rapi_rel_sql()` is still present and still internal; the first-class replacements it was to be deleted in favour of — `rel_to_table()`, `rel_to_view()`, `rel_from_table()` — now exist, so the open verdict is the removal itself), [#1147](https://github.com/duckdb/duckdb-r/issues/1147) (the seam converts only exceptions that unwind through it; anything else reaches `std::terminate()` — unreproduced, stated as a boundary), and [#1796](https://github.com/duckdb/duckdb-r/issues/1796) (raising an R error from an ALTREP method is unsafe; today the materialization paths still do it). None was closed or commented on.

## What moved

* `AGENTS.md` — lost the bodies of §§ &#34;C++ Glue Code Conventions&#34; and &#34;C++ Warning Policy&#34;, each replaced by a one-line pointer to the leaf; gained the visible italic handbook backreference under its H1. Nothing else was touched, since five other leaves are splitting this file in parallel.
* `src/include/rapi.hpp` — gained a two-line plain source comment (`// Handbook: handbook/architecture/glue/`) directly after `#pragma once`, as the backreference for the glue sources.

## Assumptions

* **One backreference for `src/`, not one per unit.** The rules say a source file with no generated index covering it carries its own comment; `src/` has no index. I put the comment only in `src/include/rapi.hpp`, the header the glue units share, rather than in every hand-written `.cpp`, to avoid a wall of near-identical comments. If per-file backreferences are wanted, that is a follow-up.
* **`AGENTS.md` § &#34;Format Checking (Optional - Has Known Issues)&#34; was left in place.** It is formatting content and therefore arguably mine, but it is a `###` under &#34;Working Effectively&#34; and not one of the two headings my work order named, so I did not touch it; the leaf states the known-issue fact independently rather than deleting it there. Whoever owns &#34;Working Effectively&#34; should point it at this leaf.
* **`make format-check` verified by running it.** Without `cmake-format` installed it does not report a difference — it raises `FileNotFoundError` on `src/CMakeLists.txt` and exits non-zero. The glue `.cpp`/`.hpp` files themselves pass. The leaf says exactly that; `AGENTS.md`&#39;s weaker &#34;currently shows formatting differences&#34; is what I replaced.
* **`AGENTS.md` says `src/` holds &#34;~30 files&#34;; that is wrong today.** `src/include/glue.mk` and `ls src/*.cpp` agree with each other and not with `AGENTS.md`. Rather than trade one number for another, the leaf names `glue.mk` as the roster. The stale &#34;~30&#34; also appears in the fast-path sections of `AGENTS.md`, which belong to another leaf — I did not sweep it.
* **The `--sort-includes=0` flag is described, not explained.** I could not verify *why* include sorting is disabled (it is inherited from upstream DuckDB&#39;s `format.py`), so the leaf states only what the flag does.
* **`scripts/flavor.patch` re-verified.** Its `src/include/rapi.hpp` hunk already applied with `fuzz 2` on `main`; after my comment it applies with `fuzz 2` and a 4-line offset. No regression, but a reviewer may prefer the patch context refreshed.

## Durability sweep

**Out:**

- *"fifteen glue translation units"* and *"names exactly those fifteen as `GLUE`"* → the mechanism: `src/include/glue.mk` is the roster, its `GLUE` variable names them all. This was the leaf's most load-bearing hostage — adding one glue file falsified it in two places.
- *"thirty-seven `rapi_rel_*` and `rapi_expr_*` entry points"* → the prefixes alone. The point of that table row is what `relational.cpp` does, not how many functions do it.
- *"sixty-four of them today"* (entry points) → the mechanism: `src/cpp11.cpp` and `R/cpp11.R`, both generated from the `[[cpp11::register]]` annotations, are the current list.
- *"Ten of the fifteen translation units call it"* → *"Most of the glue calls it"*. The argument is about passing `context` accurately; the ratio only supports it.

**Kept:**

- The **two disjoint bodies of code** in `src/`, the **two ways an error crosses** the seam, the **two replacement scans**, and the **two `duckdb.materialize_*` options** — each names its members, and each number is the design rather than an inventory of today's tree.
- The **`make format-check` / `cmake-format` behavior**, which is a verified behavior, not a count.
- *"no glue translation unit contains such a pragma today"* — an invariant the stated policy enforces, not something a routine commit is supposed to change.
- The per-unit responsibility table, which is an inventory rather than a count.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,7 @@
 # DuckDB R Package - Operational Instructions
 
+*Handbook: [`architecture/glue/`](/handbook/architecture/glue/README.md).*
+
 R package that contains a vendored copy of the DuckDB C++ library and glue code for R, including a DBI and a relational interface.
 
 ## Where to look
@@ -236,9 +238,7 @@ R
 
 ## C++ Warning Policy
 
-- **Do not suppress warnings with `#pragma clang diagnostic ignored` or similar.** CRAN rejects packages that silence warnings rather than fixing the underlying issue.
-- Fix the root cause instead. For vendored code in `src/duckdb/`, add a patch file in `patch/` that corrects the source of the warning (e.g. by changing template definitions to avoid instantiating deprecated types).
-- Example: `-Wdeprecated-declarations` from `char_traits<T>` for non-char `T` in libc++ was fixed by changing the `std_string_view` alias in `src/duckdb/third_party/fmt/include/fmt/core.h` to a struct that only provides `std::basic_string_view<Char>` for standard char types.
+See [`architecture/glue/`](/handbook/architecture/glue/README.md).
 
 ## Never Hard-Code the Package Name
 
@@ -279,8 +279,7 @@ and CI fails at the `roxygenize` step, before anything is compiled.
 
 ## C++ Glue Code Conventions
 
-- R string constants and symbols (SEXP) used in C++ glue code are defined in `src/utils.cpp` (in `RStrings::RStrings()`) and declared in `src/include/rapi.hpp` (in `struct RStrings`).
-- Always add new string constants and `Rf_install()` symbols to `RStrings` rather than using inline `StringsToSexp()`, `Rf_mkString()`, or `Rf_install()` calls in hot paths.
+See [`architecture/glue/`](/handbook/architecture/glue/README.md).
 
 ## Dependencies
 
--- a/handbook/architecture/glue/README.md
+++ b/handbook/architecture/glue/README.md
@@ -1,25 +1,246 @@
 # The C++ glue
 
-*Stub — this leaf will own its topic;
-today it routes to where the knowledge lives.
-The writing protocol is in [`meta/handbook/`](/handbook/meta/handbook/);
-the last section holds this leaf's parameters.*
+The translation units in `src/` that bridge R and DuckDB —
+the cpp11 entry points, the cached R constants in `RStrings`,
+the ALTREP relations, and the path an error takes back into R —
+together with the conventions that govern C++ source in this package.
+The vendored engine underneath is
+[`architecture/engine/`](/handbook/architecture/engine/README.md);
+the R code on the other side of the seam is
+[`architecture/r-layer/`](/handbook/architecture/r-layer/README.md).
 
-Scope: the translation units bridging R and DuckDB —
-cpp11, `RStrings`, ALTREP relations, error rethrow —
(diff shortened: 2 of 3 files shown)


krlmlr:
`main` now carries a written leaf for the C++ glue at core depth (#2493), so this branch predates the tree it edits, conflicts throughout, and merging it would revert the newer leaf. Closing unmerged.

Salvaged from here: that `make format-check` aborts with a `FileNotFoundError` on `src/CMakeLists.txt` when `cmake-format` is absent, instead of reporting a difference — actually run, not inferred; and that the ALTREP materialization budget pushes a `LimitRelation` of the budget plus one row and raises when that row comes back, so an over-budget relation fails rather than arriving silently truncated, under a temporarily doubled `max_expression_depth` (#101).

Both are recorded in `plan/history/2026-08-handbook-wave-salvage.md` — #2507, [file](https://github.com/duckdb/duckdb-r/blob/claude/handbook-wave-salvage-zayxqn/plan/history/2026-08-handbook-wave-salvage.md).

---

---

PR #2485, opened 2026-08-02
docs(handbook): Write the per-commit builds leaf

krlmlr:
## What this leaf now owns

`handbook/operations/ci/per-commit/` is the single explanation of the sharded per-commit build system: the three jobs of `each-rcc` (`plan` → `build` → `harvest`), how the range is bounded by `{S}-green` and how a `retry-{S}-dev` branch replans its tip, the gate list and the frozen `snapshot-{sha}-rcc-smoke-null` branch name, the commit status as a display-only surface, the orphan `rcc` branch as the verdict store (`runs2.d/` as write surface, `runs2.ndjson` as read surface, publishing cost, the measured race, newer-verdict-wins, consolidation), the cost model with its four measured constants and both partition passes, the operating knobs, how to read a failure from the job summary without opening the log, and the failure-mode table. Boundaries are linked rather than restated: the vendoring loop that feeds it, the green marker as a series guarantee, platforms and R versions, the workflow inventory, and the operator's view of a red run.

(Refs and branch names are written with braces here — `{S}-green`, `retry-{S}-dev`, `snapshot-{sha}-rcc-smoke-null`. The leaf uses angle brackets; GitHub deletes those from a pull-request body even inside backticks.)

## What moved

- **`scripts/EACH.md` — deleted.** Fully absorbed; no tombstone. `scripts/README.md` is the in-place index that routes on in its place.
- **`.claude/skills/docs-consistency/docs-readme.R`** — lost the now-dead `EACH.md` glob from this leaf's `globs =` group; `each-*` and `rcc-*` still map here.
- **`scripts/README.md`** — regenerated; lost its `EACH.md` row, still routes the `each-*`/`rcc-*` scripts to this leaf.
- **`.github/workflows/each.yaml`, `rcc-logs.yaml`, `rcc-consolidate.yaml`** — gained a `# Handbook: handbook/operations/ci/per-commit/` comment; `each.yaml`'s two prose pointers now name the leaf instead of the deleted file.
- **`scripts/rcc-decided.sh`, `each-plan.sh`, `rcc-part-push.sh`, `rcc-parts-test.sh`** — comments repointed from `scripts/EACH.md section 3` to the leaf.
- **`AGENTS.md`** — its routing row for per-commit CI now names `handbook/operations/ci/per-commit/`.
- **`BRANCHES.md`** — lost the `EACH.md` line from its directory tree; two pointers repointed to the leaf.
- **`scripts/VENDORING.md`** — three `EACH.md` links repointed to the leaf.
- **`plan/README.md`, `plan/history/vendoring-loop.md`, `.claude/skills/series-loop.md`** — links repointed to the leaf.
- **`handbook/operations/vendoring/troubleshooting/README.md`** — its stub's `Today:` list no longer links `/scripts/EACH.md`; it now names this leaf as what absorbed the file. See the assumption below, which this reverses.

## Assumptions

- **Where `EACH.md` and the code disagreed, I took the code and said so in the leaf.** Six corrections: the object count in the constants table was wrong — `src/include/sources.mk` does not list the 355 objects it claimed, verified by parsing the file (the leaf now names the file rather than any count; see the durability sweep below); **`PENDING_TTL_HOURS` is implemented nowhere** in `scripts/` or `.github/`, so the leaf says a commit with no record is simply undecided and a `pending` status is safe to disbelieve, rather than carrying the retired knob forward into the durability and failure-mode tables; the **resume check reads the verdict store** (`each-shard.sh` → `rcc-decided.sh`), not "one REST read per commit"; the **effective `max-parallel` default is 20** (the workflow always passes a value) even though `each-plan.sh`'s own default is 8 — the leaf states both; **`force` is a `workflow_dispatch` input only**, not exposed on `workflow_call`; and **every job is gated `if: github.repository == 'krlmlr/duckdb-r'`**, which `EACH.md` never mentions and which means the workflow is inert in this repository and in forks. Please confirm the last one is intended and not a leftover.
- **Length.** The work order expected well under a third of the 1237-line source; the leaf lands at ~42% fewer words (6.2k against 10.7k). I cut the argument and the design-document rehearsal, but the work order also pinned the measured constants, the failure modes, and the failure-reading procedure as must-keeps, and the handbook rules require a leaf to state its limits and declined requests — between them those account for most of what remains. I chose fact-preservation over the length target; a reviewer who disagrees should say which facts to drop.
- ~~**One dangling link is left deliberately.** `handbook/operations/vendoring/troubleshooting/README.md` still lists `/scripts/EACH.md` under its stub's `Today:`. That is another leaf's `README.md`, being written in a parallel branch that rewrites the file wholesale, and scope discipline says not to touch it. It resolves on merge.~~ **Reversed.** A branch must not carry a dangling link on its own, whatever the merge will do to it, so the one line was repointed at this leaf. The parallel branch (#2472) rewrites that file wholesale and its version supersedes this edit; expect a conflict there and resolve in favour of #2472.
- **`plan/PLAN-vendoring-simplification.md` was left alone.** It mentions `EACH.md` in backticks (never as a link, so nothing dangles), including a docs-inventory row calling it "the single owner of its topic", which is now stale. It is a plan document recording intent at plan time, and editing its inventory belongs to whichever leaf absorbs that plan.
- **#2478 also edits `scripts/EACH.md`**, repointing two `BRANCHES.md` anchor citations at `branches/invariants/`. Those edits vanish with the file; nothing else there depends on it surviving.
- The numbers I could not recheck cheaply — the fitted constants and their residuals, the race distribution, the consolidation before/after, the publish timings, the backlog planning figures — are kept as the measurements the source reports them to be, and the leaf says they are measured. Everything structural (gate list and order, env vars and defaults, triggers, concurrency groups, permissions, timeouts, retired scripts and workflows) was verified against the tree.
- No issues were listed for this leaf, so none were drained.

## Durability sweep

A follow-up pass separated the measurements this design rests on from the
counts that merely describe today's tree. The measurements stay, and each
now says it is one; the descriptions are gone.

**Out**, replaced by the mechanism or a qualitative form:

- "341 objects" in three places → "every object `src/include/sources.mk`
  lists", "the object list from `src/include/sources.mk`", "the unity
  objects `src/include/sources.mk` lists". A reader who needs the number
  reads it there, correct on the day they read it. That this count had
  already been corrected once against its source is the tell.
- "138 of them `ub_*.o` groups" → "many of them"; "3517 files" → gone.
  Neither carried an argument.
- "about 3.5 seconds" for `each-cost.py` → "a few seconds".
- "applies six gates" → "applies its gates"; the list follows immediately.
- "About 950 of those branches exist" and "~950 snapshot branches" →
  "those branches accumulate" and "the accumulated snapshot branches".
  The argument is that tooling looks them up by name.
- "around 218 MB" → "hundreds of megabytes"; "all 2.5k logs — the whole
  220 MB" → "every log on the branch — the whole of it". (218 MB still
  appears once, inside the consolidation before/after table, which is
  framed as measured — see below.)

**Framed rather than deleted**, since a measurement with provenance ages
honestly where a bare number pretends to be current:

- The ccache figure now carries its source: measured over eight
  consecutive `v1.5` commits, recorded in Appendix A of the superseded
  vendoring-loop design.
- "Peak RSS per unity object is about 834 MB" → "measured about".
- The reach-map sanity table is "measured against the vendored tree as it
  stood then"; the consolidation table is measured on the live branch "as
  it stood when the script was written".

**Kept:** the fitted constants and their defaults, the RMSE and the
residual table, the three cold builds, the `-j2`/`-j4` timing, the
object-count distribution over 120 `v1.5-variegata-dev` commits, the
planner runs on named backlogs, and the publish-race distribution — each
already names the population it was measured over. Also kept the
load-bearing counts (three jobs, two passes, the 256-way fan-out, two
commits after consolidation, GitHub's five limits), the config defaults,
and the `max-parallel` discrepancy between the workflow and
`each-plan.sh`, which is a boundary the leaf documents on purpose rather
than a stale count.

---


krlmlr:
`main` now carries a whole subtree for per-commit CI (#2493, #2504) — contract, legs, operating, planning, selection, store — which is deeper than this single page and states most of what it states, the split-factor table and the `-j2` against `-j4` measurement among them. This branch predates that, conflicts throughout, and merging it would revert the subtree. Closing unmerged.

Salvaged from here: the memory headroom behind `-j4`. Peak RSS measured about 834 MB per vendored unity object, so four compiles in flight fit a 16 GB runner, and nothing else oversubscribes — no `-flto` is requested, R's `Makeconf` reports `LTO =` empty, and `DESCRIPTION` carries no `Config/testthat/parallel`, so the suite runs serially. The three negatives are what make the measurement actionable.

Not salvaged: the `PENDING_TTL_HOURS` observation, which the subtree now states in its durable form — `pending` is a display artifact and selection reads records.

---

---

PR #2487, opened 2026-08-02
docs(handbook): Let internal nodes carry their area's governing principles

krlmlr:
Amends the rule that said an internal `README.md` may carry no prose, so that
it may state the principles governing its area — and applies it across the
tree. Five nodes gained principles; seven were examined and deliberately left
navigation-only.

## The line

> A **principle** is a constraint the area places on its leaves — why the area
> is divided as it is, or a rule every leaf under it obeys. Its truth-maker is
> the set of children, not any one of them. A **summary** is a conclusion one
> leaf reached, repeated a level up; its truth-maker is that leaf.

What the rule ships is the operational form — three tests a sentence on an
internal node must survive:

1. **A leaf yet to be written could falsify it.** A statement one leaf settles
   *in full* is that leaf's, however tree-wide it feels: nothing a later leaf
   says could make it false, which is what tells you it is finished and owned.
   A statement each leaf settles only for itself, and which holds of the area
   only because all of them do, is the node's. **A conjunction of finished
   claims is a summary; a generalisation over instances is a principle.**
2. **It has the lifetime of the child list.** An internal node may go stale
   only when its children change — which is when the node is being edited
   anyway. A sentence an ordinary commit to the package could falsify is a
   fact, and every fact has a leaf.
3. **It names no particulars.** No paths, script names, environment variables,
   versions, counts, commands, or tables. That is the vocabulary of facts.

Test 2 is why this does not reintroduce the rot the navigate-only rule was
built against: the node's text and its child list have the same lifetime, so
the node goes stale exactly when someone is already editing it.

Test 1 is in its sharp form because review found the blunt form insufficient —
see [What the review changed](#what-the-review-changed) below. It is the test
that does the real work.

Two enforcement handles, both in the rule text:

**The link rule.** An internal node links to its children, and to another
area's node where it draws a boundary. It never links *into* a leaf — that is
the node reaching for the leaf's evidence, which means the sentence was the
leaf's all along. Mechanically checkable, with exactly one allowlisted edge
(the root's pointer to the rules page), named in the rule so the check stays a
check.

**Look for an owner first.** Where an area has a leaf whose topic is the area's
own reasons or its own rules, the principles are already that leaf's and the
node stays navigation-only. This decided three areas.

## Nodes that gained principles

| Node | Principle |
|---|---|
| `handbook/` | The areas divide by what a question is about, not by who is asking it |
| `usage/` | The package decides as little as it can, and says so when it must |
| `architecture/` | Little of what ships was typed where it is read |
| `build/` | What ships is the plain build; a shortcut fails loudly or not at all |
| `testing/` | Nothing here passes by being silent |
| `operations/releases/` | A release asserts what it is, it derives nothing; and the rules are other people's |

Each also carries the rationale for how its area is divided, which is licensed
by the child list rather than by leaf content — so it is safe in areas whose
leaves are still stubs. The root now carries *only* that rationale.

Each surviving principle passes test 1 in the sharp form: a leaf added tomorrow
could falsify it. A `build/` leaf documenting a shortcut that fails quietly, a
`testing/` leaf whose check skips in silence, a `usage/` leaf with a silent
default, an `architecture/` leaf that is a large hand-written subsystem, a
`releases/` leaf with a derived version — each would break its node, which is
what makes the node's sentence a claim about the area rather than a report from
below.

**`build/`** is the cleanest case: two principles, each instantiated in all
three leaves and stated as a rule by none. **`testing/`** carries the boundary
against `operations/ci/`, the sanctioned kind of outward link.

## Nodes deliberately left navigation-only

An area with no principle getting none is the rule working, not a gap in it.

* **`operations/vendoring/`** — the candidate was "git is the only state", and
  it fails test 1: `model/` already owns "the vendor commit is the record" and
  `series-loop/` already says it keeps no state of its own. Two finished claims;
  the node would add only their conjunction. Structurally, `model/` is the
  area's rationale leaf, and a rationale leaf absorbs the node's principles by
  construction.
* **`branches/`** — same verdict, sharper reason. `invariants/` is a leaf whose
  entire topic is the area's governing rules, numbered, with what would catch
  each violation. Anything the node said would summarise it.
* **`operations/`** — five heterogeneous children; nothing binds them beyond
  the scope sentence it already has. A node high in a heterogeneous area is
  often only a junction.
* **`operations/ci/`** — only `per-commit/` is written; `workflows/` and
  `matrix/` are stubs. A principle asserted against one written leaf of three
  is a guess.
* **`contributors/`** — three leaves in temporal order. Ordering is navigation.
* **`meta/`** — the candidate is `meta/plans/`'s own subject and is stated there.

## What the review changed

The root originally carried three principles. Review caught the first, and
re-testing the other two against the same standard dropped both.

**"No page here may assume the package is called `duckdb`"** was two things
under one heading. The bolded sentence is a rule about *pages* — a
documentation-writing rule. The paragraph under it was a fact about the
codebase: one source tree, several names, applied per branch. They have gone to
different places.

The writing rule is now in `meta/handbook/`, where page rules live, phrased as
what it is: no page writes `duckdb` as though it were the package's name,
because a rename falsifies it everywhere but the mainline flavor. It is worth
having written down because it is **mechanically checkable and nothing else
checks it** — the scan that catches a hard-coded name in the shipped sources
does not reach `handbook/`, and names that as one of its own limits. Three
spellings are called out as correct so the check does not produce them as false
positives: the repository (`duckdb-r`), the engine and upstream project (not
ours), and a flavor named as a flavor.

The fact left the root entirely. `branches/flavors/` owns it, and the
navigation list already names flavors under both `usage/` and `branches/`.

**"The engine is carried in this repository, not depended on"** — dropped.
`architecture/engine/` settles it in full; nothing a later leaf could say would
make it false. Finished and owned, so a summary.

**"Every commit is built and tested on its own"** — dropped. `branches/invariants/`
is a leaf whose topic is the area's own rules and states this as one of them,
which is precisely the "look for an owner first" clause I had already derived
for `operations/vendoring/` and `branches/`. The clause kills this one too.

The root keeps its division rationale, which no leaf can settle and which a
ninth area could falsify. It is a real constraint on the tree's shape — an
audience split would need the same fact in two places, and the tree holds every
fact once — and it passes cleanly.

The sharp form of test 1 is what this review bought. The blunt form ("cannot be
checked by reading one leaf") let the flavor fact through, because the rule
stapled to it *was* genuinely tree-wide. The falsifiability form catches it: a
new leaf could not make "one source tree is published under several names"
false, so it was already owned. The same form independently confirmed both
other drops, and confirmed all six survivors.

## `meta/handbook/`: keep one page

Unchanged; the maintainer has not asked to revisit it. The rejected alternative
was `meta/handbook/{tree,leaf,forms,enforcement}/`, one sub-leaf per `##`.
Chapters are not topics; the material is cross-referential and read as a unit;
and it fails the proposal's own test, since as an internal node it could carry
only principles governing its children, whose content *is* the rules. **A node
whose principles equal its children's content should not be split.** Future
trigger: split when a section becomes separately maintained, not when the page
gets long — likeliest candidate "Writing a leaf", whose home would be
`.claude/skills/`.

## What could still go stale

Test 1 and the link rule are mechanical; test 2 and test 3 are judgement.

The standing risk is a node that is right when written and becomes a summary
later without anyone touching it — an area acquires a leaf that settles the
node's principle in full, and the node is now redundant with a leaf it predates.
Nothing detects that. Adding a leaf edits the child list, which is when the rule
puts the node up for review, and "look for an owner first" tells the writer
where to check.

The review moved one risk and left one. **Moved:** the root is no longer
carrying facts with owners, which was the live instance of that failure and had
already shipped. **Left, and now the largest:** `usage/`'s principle rests on
three written leaves out of eight. Its division rationale is safe — that follows
from the child list — but "announced, reversible, or refused" is as much a
prediction about `extensions/`, `memory/` and `types/` as a reading of
`storage/` and `integrations/`. By test 1 that exposure is what makes it a
principle rather than a summary, so this is the rule working as intended; but it
means the node is what changes if a stub lands that contradicts it, and it is
the one to check first when they do.

Verified: no dangling links, no upward links, root-absolute for everything
leaving its directory, semantic line breaks throughout. Every interior node
links only to its children, plus `testing/` → `operations/ci/` and the one
allowlisted root edge. No node written here writes the package's name. No leaf
`README.md` was touched.

### DIFF
--- a/handbook/README.md
+++ b/handbook/README.md
@@ -5,28 +5,16 @@ for every documentable aspect of this package;
 internal pages like this one navigate and state their area's principles,
 leaves explain ([the rules](meta/handbook/)).
 
-Three constraints run through every area below,
-and a reader who does not carry them will misread all of them.
-
-**No page here may assume the package is called `duckdb`.**
-One source tree is published under several names,
-applied per branch rather than chosen at build time,
-so the name is a question asked at run time wherever it appears —
-in the R code, in the tests, in what a user installs.
-
-**The engine is carried in this repository, not depended on.**
-A question about DuckDB and a question about the R package
-are therefore answered in the same tree, at different leaves,
-and drawing that boundary is part of what every area does.
-
-**Every commit is built and tested on its own.**
-History advances one upstream commit at a time and stays linear,
-which is what makes a regression attributable;
-much of the machinery described below exists to keep that affordable,
-and nothing else explains the shape it has.
-
-The areas divide by what a question is about, not by who is asking it:
-a user and a maintainer with the same question arrive at the same leaf.
+The areas divide by what a question is about,
+not by who is asking it.
+A reader who will only ever install the package
+and one who maintains the repository that produces it
+are served by the same tree,
+and the same question brings them both to the same leaf.
+That is why the list below mixes using, building and operating
+rather than separating them by audience:
+an audience split would need the same fact in two places,
+and the tree holds every fact once.
 
 * [`usage/`](usage/) — installation and flavors, connections,
   types, extensions, memory, data import, storage, integrations
--- a/handbook/README.md
+++ b/handbook/README.md
@@ -2,8 +2,31 @@
 
 The single source of truth
 for every documentable aspect of this package;
-internal pages like this one only navigate — leaves explain
-([the rules](meta/handbook/)).
+internal pages like this one navigate and state their area's principles,
+leaves explain ([the rules](meta/handbook/)).
+
+Three constraints run through every area below,
+and a reader who does not carry them will misread all of them.
+
+**No page here may assume the package is called `duckdb`.**
+One source tree is published under several names,
+applied per branch rather than chosen at build time,
+so the name is a question asked at run time wherever it appears —
+in the R code, in the tests, in what a user installs.
+
+**The engine is carried in this repository, not depended on.**
+A question about DuckDB and a question about the R package
+are therefore answered in the same tree, at different leaves,
+and drawing that boundary is part of what every area does.
+
+**Every commit is built and tested on its own.**
+History advances one upstream commit at a time and stays linear,
+which is what makes a regression attributable;
+much of the machinery described below exists to keep that affordable,
+and nothing else explains the shape it has.
+
+The areas divide by what a question is about, not by who is asking it:
+a user and a maintainer with the same question arrive at the same leaf.
 
 * [`usage/`](usage/) — installation and flavors, connections,
   types, extensions, memory, data import, storage, integrations
--- a/handbook/architecture/README.md
+++ b/handbook/architecture/README.md
@@ -2,6 +2,22 @@
 
 What the shipped code is:
 the R layer, the C++ glue, and the vendored engine.
+The area divides by who wrote the code rather than by what language it is in:
+two of the three are written here,
+and the third is upstream's, described only as far as
+this package's build of it differs from a stock one.
+
+**Little of what ships was typed where it is read.**
+A source here is as likely to be generated from an annotation,
+derived from a sibling file, or copied wholesale from upstream
+as it is to be hand-written,
+so the first thing a leaf says about a file is which of those it is.
+The consequence is a habit rather than a rule:
+a correction goes to the producer, never to the file in front of you,
+and a page in this area earns its keep by naming the producer
+rather than by describing the product.
+The hand-written surface is smaller than the directory listing suggests,
+and knowing which part of it is hand-written is most of what this area is for.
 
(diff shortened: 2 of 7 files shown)


krlmlr:
`main` now carries principles on its internal nodes, and #2504 added the three tests a principle has to pass before it may sit on one. This branch predates that, conflicts throughout, and merging it would revert the newer tree — including several of the principles it proposed, in their tested form. Closing unmerged.

Nothing was salvaged from here. Principles are judgements about the tree rather than claims about the code, so there was nothing to verify; what the branch was reaching for now exists, with a rule for when a new one may be added.

---

---

PR #2492, opened 2026-08-02
fix(patch): Fix the re2 clang warnings instead of hiding them

krlmlr:
`patch/0003` suppressed three clang warnings in re2 with
`#  pragma clang diagnostic ignored`. That spelling falls outside the
`^\s*#pragma (GCC|clang) diagnostic ignored` regex `tools:::.check_pragmas`
greps for, so `R CMD check` never reported them while the compiler
honoured them in full.

Renamed to `patch/0003-Fix-clang-warnings-in-re2.patch`, and it now fixes
rather than silences.

> **Scope note.** This was originally a re2 *and* mbedtls change; the
> mbedtls half has been dropped at the maintainer's request, so
> `patch/0016` and `constant_time_impl.h` are untouched here. The `-Wvla`
> question that half raised is not decided by this PR.

## What changed

* **`-Wnested-anon-types`** — two unnamed structs declared inside
  anonymous unions in `prog.h`, now named and hoisted out as
  `Inst::ByteRange` and `Prog::PrefixFrontBack`. Naming them in place is
  an *error* ("types cannot be declared in an anonymous union"), hence
  the hoist. Layout-identical: `sizeof(Prog)` and alignment unchanged,
  `static_assert(std::is_trivial<Inst>::value)` still holds, and a linked
  `RE2::FullMatch` / `PartialMatch` smoke test passes.
* **`-Wdtor-name`** — `Regexp::Walker<T>::~Walker()` in `walker-inl.h`,
  now `Regexp::Walker<T>::Walker::~Walker()`, which is clang's own
  fix-it. It fired in every translation unit including that header.
* **`-Wgnu-anonymous-struct`** — never fired here, because this copy
  names its union members. Suppression dropped, not replaced.

The pragmas were also costing a warning of their own: GCC does not know
`#pragma clang`, so each raised `-Wunknown-pragmas` per translation unit
under `-Wall`. Every re2 translation unit is now warning-free under both
compilers.

## Which flags surface these

Measured on minimal reproductions of each construct, clang 18.1.3 and
g++ 13.3, Linux x86-64:

| construct | clang `-Wall` | clang `-Wall -pedantic` | g++ `-Wall -pedantic` |
|---|---|---|---|
| named union member, nested type | — | `-Wnested-anon-types` | — |
| unnamed union member, nested type | — | `-Wgnu-anonymous-struct`, `-Wnested-anon-types` | `-Wpedantic` |
| out-of-line template dtor | — | `-Wdtor-name` | — |

So **`-pedantic` is what surfaces all three, and only clang reports
them** in the form this copy has. `-Wall` alone is silent, which is why
nothing noticed. Companion PR #2508 wires a check accordingly.

## Would this apply upstream to `google/re2`?

Both defects are still present on `google/re2@main` —
the anonymous nested struct at `re2/prog.h:175`, the destructor at
`re2/walker-inl.h:142` — so the fixes are wanted there. But the patch
file does not transplant cleanly. Tested against a fresh checkout of
those two files:

* `-p5` is the right strip level (it removes
  `a/src/duckdb/third_party/re2/`). `-p4` finds nothing.
* **2 of 4 hunks apply**: the `ByteRange` type definition (offset 27) and
  the whole `walker-inl.h` destructor fix (offset 2).
* **2 hunks fail, and `-F3` does not rescue them**, because both
  mismatches are interior rather than at the hunk edges:
  * the hunk replacing the byte-range member carries the context line
    `Match ID to identify this match (for duckdb_re2::Set)` — a
    vendoring rename; upstream reads `re2::Set`.
  * the `prefix_front_back` hunk targets `} prefix_front_back;`, a
    *named* member. Upstream leaves that member **unnamed**.

That last difference matters: upstream's unnamed member is the stronger
case. Per the table above it additionally triggers
`-Wgnu-anonymous-struct` — the warning dropped here as never-firing —
and it is the one construct g++ also flags. An upstream patch would need
to name the member as well as the type, so it is a rewrite of two hunks
rather than a port.

## Verification

* Reverse-applied all 14 patches to recover a pristine `src/duckdb/`,
  re-applied the old stack, and diffed byte-for-byte against the
  worktree, to establish the baseline before changing anything.
* Forward-applied the new stack over pristine: all 14 apply
  `--forward --dry-run` clean, in filename order.
* `tools:::.check_pragmas(".")` returns `character(0)` before and after,
  so `R CMD check` output is unchanged by this PR.
* Not run: `R CMD check`, a full package build. Compiler evidence is
  Linux x86-64 clang 18.1.3 and g++ 13.3 only; the CRAN Fedora-clang and
  Rtools-GCC flavours were not exercised.

## Still to do before merge

`REVERT ME: Temporarily put back old code to see if our checks are good`
is still the branch tip, deliberately, so #2508's CI run goes red on the
defect this PR removes. **It must be dropped before merging**, and this
PR merges before #2508.

---


krlmlr:
Also checking in winbuilder.

krlmlr:
Checks now expected to be green. Can be merged then.

PR #2499, opened 2026-08-03
fix: avoid strict normalization for new database paths

### LINKED ISSUE #455: Creating new databases fails on network drives if user has no privileges on underlying folder due to `normalizePath` with `mustWork = TRUE`
Recently, in https://github.com/duckdb/duckdb-r/commit/27bba58e12d265527e0c64da97dd3ab63d2b2c8a, the R package for duckdb has been updated to normalize paths before connecting, using `normalizePath` with `mustWork = TRUE`

However, `normalizePath` will fail on network drives if the user does not have sufficient permissions on all underlying directories, even when the path is valid and the user can access it. This causes existing code to break with the new package version.

Quoting from the docs of `normalizePath`:

> Converting to an absolute file path can fail for a large number of reasons. The most common are
> -  ...
> - A component before the last is not a directory, or there is insufficient permission to read the directory.
> - ...

I can work around this by monkeypatching the package, e.g.

``` r
unloadNamespace("duckdb")
path_normalize_noerror <- function(path) {
  DBDIR_MEMORY <- ":memory:"
  if (path == "" || path == DBDIR_MEMORY) {
    return(DBDIR_MEMORY)
  }

  out <- normalizePath(path, mustWork = FALSE)

  # Stable results are only guaranteed if the file exists
  if (!file.exists(out)) {
    on.exit(unlink(out))
    writeLines(character(), out)
    out <- normalizePath(out, mustWork = FALSE)
  }
  out
}
assignInNamespace("path_normalize", path_normalize_noerror, "duckdb")
```

However, I'd rather not do any monkeypatching. After monkeypatching, everything works fine.

Note that this operation is safe in that either `file.exists(out)` is `TRUE` and thus `normalizePath(out, mustWork = TRUE)` would never be called, or that `writeLines(character(), out)` is called and thus the path must be valid (as a file can be created on that path).

I'm not entirely sure why `mustWork = TRUE` was used there. I'd gladly submit a pull request changing that to `FALSE` if there's no good reason for it.

LeonidasZhak:
## Summary

Keep both `normalizePath()` calls non-strict when opening a new database path. Fixes #455.

## Thanks to maintainers

Thanks for the guidance on #455 and for reviewing this path-handling edge case.

## Issue or motivation

On some network drives, strict normalization can still fail immediately after the placeholder file is created. That prevents DuckDB from opening an otherwise writable new database path.

## Root cause

`path_normalize()` first used `mustWork = FALSE`, created a placeholder for a stable result, then called `normalizePath(..., mustWork = TRUE)`. The second call reintroduced the network-drive failure the first call was intended to avoid.

## Change

Use non-strict normalization on the second call as well, and check explicitly that the placeholder was created. The injectable normalizer is only a test seam for verifying both calls.

## Tests

- Focused path-normalization test: 3 passed.
- Path, connection, and read-only tests: 10 passed, 10 conditionally skipped.
- Vendored-source package install and a file-backed database smoke test passed.
- `R CMD check --as-cran --no-manual`: package installation, examples, and `testthat.R` passed. The remaining NOTE is CRAN incoming metadata for the development version.

## Scope

This does not change existing-file handling, extension-prefixed database identifiers such as `md:`, URL handling, or connection APIs.


krlmlr:
Thank you.

I've noticed that you've opened hundreds of PRs across the ecosystem. What is the project behind this?

LeonidasZhak:
Thanks for asking. This is a personal project: I'm trying to make steady, useful contributions to the R ecosystem while deepening my understanding of R codebases and package development. It isn't affiliated with any organization or commercial service.

I did open too many PRs too quickly earlier, and I understand why that raised concern. I've since adjusted the process to at most one carefully scoped contribution per day, prioritizing follow-up on existing PRs and requiring local package checks before submission. I use AI assistance to help inspect code, refine changes, and run tests, but I review each patch and remain responsible for it.

For this PR, I followed #455 and your earlier comment that strict normalization should be replaced with a manual existence check. If duckdb-r has any contributor preferences I should follow, I'm happy to respect them.

LeonidasZhak:
Thanks for asking. The common thread is a personal, agent-assisted workflow for contributing narrowly scoped fixes and documentation to open-source R projects. Each PR is tied to a concrete issue and tested where practical. I’m happy to follow duckdb-r’s contribution and AI-disclosure guidelines; please let me know if you’d prefer any project-specific disclosure or workflow.

PR #2508, opened 2026-08-03
feat(ci): Guard the vendored engine against silenced warnings

krlmlr:
Companion to #2492, answering the review finding there:
*"CI can't detect the failure mode that the patches resolve.
Do we need to adapt CI?"*

**Yes.** Nothing in CI could see it, in either of the two ways it hides.
This adds `scripts/vendored-warnings.sh` and runs it in the smoke test.

**Merge order: #2492 first, then this.**
This branch is built on #2492's, so the order is forced;
merged the other way round, `main`'s CI goes red on the very defect
#2492 removes (demonstrated below).
The `REVERT ME` commit at #2492's tip is still in this branch's history —
it is what makes the red build visible here,
and it must be dropped from #2492 before either merges.

## The failure mode, in two halves

`tools:::.check_pragmas()` — what `R CMD check` runs — greps `src/` and
`inst/include/` for `^\s*#pragma (GCC|clang) diagnostic ignored`:
single spaces, and no space between `#` and `pragma`.
The preprocessor is far more permissive.

1. **A suppression the check cannot see.**
   `#  pragma clang diagnostic ignored "-Wnested-anon-types"` and
   `#pragma  GCC  diagnostic  ignored "-Wvla"` are honoured by every
   compiler and matched by nothing. This is lexical, and a lexical check
   catches it.
2. **The warnings themselves.**
   Spelling aside, no job compiled this code with the warnings enabled
   and failed on them. This is a build problem, and only a compile
   catches it.

The `REVERT ME` commit isolates the second: it restores the pre-fix re2
sources *without* restoring the pragmas, so the warnings fire in the
open. A purely lexical check is identical on both sides of that commit
and therefore cannot go red then green. So the compile half is not
optional — but it does not subsume the first half either, because a
suppression that works is a suppression a compiler never mentions.
Hence both.

## What the guard does

**Half one — the suppression inventory** (no compiler, instant).
Greps the patch stack for every `diagnostic ignored` an added line
introduces, whatever the spacing, reduces it to
`patch file` + diagnostic, and compares against an inventory the script
carries. Today that inventory has exactly one entry: mbedtls's `-Wvla`,
with the reason written beside it — it guards the `asm volatile` barrier
that stops `mbedtls_platform_zeroize()` from being optimised away, and
normalising its spelling would change what `R CMD check --as-cran`
reports, which #2492 explicitly left as the maintainer's call. A new
suppression fails the check until someone writes it down with a reason.
The escape stops being available by default and becomes a listed
decision.

**Half two — warning-free vendored libraries** (needs clang, ~15 s).
`clang++ -std=c++17 -fsyntax-only -Wall -pedantic -Werror` over re2 and
mbedtls — the libraries whose patches claim warning-freedom for them.
`-fsyntax-only`, so this costs seconds rather than a build.

Clang specifically, not the compiler `R CMD config CXX` reports:
`-Wnested-anon-types` and `-Wdtor-name` do not exist in GCC. I checked —
g++ 13.3 with `-Wall -pedantic` is **completely silent** on the
`REVERT ME` tree. A guard wired to the R toolchain compiler would have
passed it.

## Where it runs

`.github/workflows/custom/after-install/action.yml`, gated on
`github.job == 'rcc-smoke'` — the same place and the same gate as the
existing flavor-name guard, for the same reason: it reads the checkout,
so its answer is identical in every matrix entry. No new workflow.

Not a testthat test: `R CMD check` runs the suite from a built tarball
on farms that need not have clang, and CRAN must not be made to compile
re2 twice.

## The upstream-spelling problem

duckdb's own vendored `pdqsort.h`, `yyjson.cpp`/`yyjson.hpp` — and
cpp11's `R.hpp` under `inst/include/` — already use the same escape
(`#   pragma clang diagnostic ignored ...`). A naive repo-wide lexical
scan goes red on code we do not maintain.

**Handled by construction, not by an allowlist:** half one reads
`patch/`, not `src/duckdb/`. Those spellings arrive with the upstream
sources, no patch of ours adds them, so the scan never sees them.
Respelling them would mean carrying a patch against files that are
re-vendored from upstream on every sync — a permanent maintenance cost
against no defect of ours. The one exception the scan *does* see is ours
(`-Wvla`), and it is listed with its reason rather than pattern-matched
away.

The same reasoning bounds half two: only libraries a patch of ours makes
a warning claim for. zstd's `dict/` sources warn upstream under `-Wall`,
and `patch/0033-clang-macos.patch` fixes a libc++ deprecation a Linux
runner with libstdc++ cannot reach — both are named in the script as
deliberately out. A guard that starts red teaches people to ignore it.

## Demonstration — red, then green

All three states run in this container, clang 18.1.3 / g++ 13.3, Linux
x86-64.

| Tree | Half one (inventory) | Half two (compile) | Exit |
|---|---|---|---|
| `origin/main` today | **red** — four suppressions in `patch/0003`, `patch/0016` | green — the pragmas work | `1` |
| this branch (`REVERT ME` in place) | green | **red** — `-Wnested-anon-types` ×2, `-Wdtor-name`, across the re2 TUs | `1` |
| `git revert --no-commit 07a4a4981` | green | green | `0` |

The middle row is the maintainer's criterion: **red at the tip, green
once `REVERT ME` is reverted.** The top row is the complement, and the
reason for the merge order: on `main` as it stands, half one reports the
four suppressions #2492 removes while half two passes, because the
invisible pragmas do their job. Each half catches exactly what the other
misses.

Verbatim, at the tip:

```
Suppression inventory: matches (1 entries).
src/duckdb/third_party/re2/re2/prog.h:148:7: error: anonymous types declared in an anonymous union are an extension [-Werror,-Wnested-anon-types]
src/duckdb/third_party/re2/re2/walker-inl.h:140:39: error: ISO C++ requires the name after '::~' to be found in the same scope as the name before '::~' [-Werror,-Wdtor-name]
...
re2 warns under clang++ -Wall -pedantic; fix it at the source.
mbedtls: no warnings under clang++ -Wall -pedantic.
```

and with `REVERT ME` reverted:

```
Suppression inventory: matches (1 entries).
re2: no warnings under clang++ -Wall -pedantic.
mbedtls: no warnings under clang++ -Wall -pedantic.
```

The worktree was restored with `git revert --abort` after each probe;
`git status` is clean.

## What it does not catch

* Warnings in vendored libraries no patch of ours makes a claim for —
  upstream's to answer for, and including them would start the guard
  red.
* A suppression written straight into the glue (`src/*.cpp`) rather than
  through a patch. `R CMD check`'s own scan covers the canonically
  spelled ones there; an evasive one would escape both. Extending half
  one to the glue is cheap if that ever happens.
* Anything platform-specific: Linux/x86-64 with libstdc++ only, so
  Apple-clang and Rtools-GCC diagnostics stay unchecked.
* `-Werror` semantics differ slightly from a real build — `-fsyntax-only`
  skips codegen, so warnings that only the optimizer emits
  (`-Wmaybe-uninitialized` and friends) are out of reach. The
  diagnostics this guard exists for are all front-end ones.

## Also in this branch

* `origin/main` is merged in, so the diff against `main` is this change
  plus #2492's; the branch was not rebased.
* Documented in `handbook/testing/guards/`, which owns it, with the
  policy link back to `handbook/architecture/glue/` ("No warning is
  suppressed"), which now says the widened spelling is a *listed*
  exception rather than a technique. Glossary line, `testing/` child
  list widened, `scripts/README.md` regenerated through its generator's
  ownership map. Link check and index freshness check both pass.

**Not run:** `R CMD check`, a full package build, and the workflow step
on a real runner — the smoke test job's `clang++` availability on
`ubuntu-26.04` is assumed from the image contents, and the script fails
loudly with an actionable message if it is absent rather than skipping
silently.

---


krlmlr:
CI detects these kinds of failures.

PR #2540, opened 2026-08-06
fix(series-advance): Set `-build-base`, rather than advance it

krlmlr:
## What happened

The 2026-08-06 series-loop firing found `v1.4-andium-build-base`
off the buffer's line,
and `scripts/series-advance.sh` refused the whole invocation:

```
Error: build-base would not move forward
```

`exit 1` there takes stage 5 with it,
so the buffer stopped extending
over a marker nothing downstream reads.

The cause was a one-off — two `chore: Auto-update from GitHub Actions`
commits pushed 31 seconds apart onto `v1.4-andium-build`
and `v1.4-andium-build-base`, from the same parent,
leaving the two refs siblings rather than one containing the other.
That is treated here as a fluke, not as a class to design around.

## The change

`-build-base` is the one ref of the four with no consumer.
The match is recomputed from the vendored-SHA scan every stage 3,
so where the ref sat before says nothing the stage needs,
and holding it to fast-forward only buys nothing
while costing an invocation whenever something moves it.

So the stage **sets** it, forcing where the match is not a fast-forward.

The other three keep their guarantees, unchanged:
`-green` fast-forwards only —
a `-green` that will not fast-forward is still the stop-and-report case —
and `-build` and `-dev` are rewritten only by this loop,
by the rules already in the skill.

`.claude/skills/series-loop.md` gets the same account:
the branch table, the stage-3 text, the script summary,
and the invariant, which now names only `-green`.

## What was checked

* `series-advance.sh` runs to completion for all three series against the
  live repository with the change in place.
* The remote accepts a non-fast-forward push to `-build-base`:
  that is exactly how the ref was unstuck by hand during the firing
  (`05876e5f9` → `6e7dd75ce`, `--force-with-lease`).
* No-op when the ref is already at the match, so a steady-state firing
  pushes nothing.
* `scripts/README.md` regenerated; the one changed row is this script's
  own header line.

Per `series-loop.md` §7, the series was unstuck by hand in the same firing,
so nothing depended on this merging.

### DIFF
--- a/.claude/skills/series-loop.md
+++ b/.claude/skills/series-loop.md
@@ -21,7 +21,7 @@ this loop repairs in place and was proven on the `main` rewind
 | `<S>-build` | the routine, unconditionally | Every upstream first-parent commit vendored one-to-one, glue compiling at every commit. **No CI runs here** (`each.yaml` never matches `*-build`). The buffer. |
 | `<S>-dev` | the routine, force-push | `<S>-build` commits, consumed up to 100 at a time, with test/R/patch adaptations folded in as CI demands. What CI builds, commit by commit. |
 | `<S>-green` | the routine, fast-forward only | The newest commit such that every commit in `<S>-green..it` has a `success` run. What r-universe should build from. |
-| `<S>-build-base` | the routine | The `<S>-build` commit equivalent to `<S>-green` (same vendored upstream SHA). Marks how much of the buffer has been consumed and verified. |
+| `<S>-build-base` | the routine, force-push | The `<S>-build` commit equivalent to `<S>-green` (same vendored upstream SHA). Marks how much of the buffer has been consumed and verified. A marker with no consumer, recomputed every stage 3. |
 
 Equivalence between `-build` and `-dev` commits is by the
 `duckdb/duckdb@<sha>` reference in the commit subject —
@@ -89,9 +89,9 @@ plus a CUTOVER line for a forward series that has caught up —
 a suggestion for a human, stage 6),
 `scripts/series-advance.sh <S>`
 (stages 3 and 5 — fast-forwards `-green`,
-moves `-build-base` by vendored-SHA match,
+sets `-build-base` to the vendored-SHA match,
 extends `-dev` by ≤ 100;
-refuses on any failure or non-fast-forward),
+refuses on any failure, and on a `-green` that would not fast-forward),
 `scripts/series-port.sh <S>`
 (stage 4 — brings `<S>-dev` level with `main`:
 cherry-picks plus a tooling sync),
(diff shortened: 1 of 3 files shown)


krlmlr:
## Second occurrence, 34 minutes after the hand-fix — and the workflow is now identified

The next series-loop firing (2026-08-06, ~16:45 UTC) found
`v1.4-andium-build-base` off the buffer's line **again**:

```
8cff76968 2026-08-06 15:52:05 chore: Auto-update from GitHub Actions  -> v1.4-andium-build-base
6e7dd75ce                     vendor: … duckdb/duckdb@6fc5718088c6dacf7d341c27f269ba2a029c0c12
```

Same one-line diff (`Config/roxygen2/version: 8.0.0.9000` → `8.1.0.9000`),
and patch-identical to `f52ed130b` from the first occurrence:

```
$ git show f52ed130b | git patch-id --stable
574a7c09a0b4669defd1624b3cb76120679cdba0 f52ed130b…
$ git show 8cff76968 | git patch-id --stable
574a7c09a0b4669defd1624b3cb76120679cdba0 8cff76968…
```

`scripts/series-advance.sh` refused with the same
`Error: build-base would not move forward`,
and this firing unstuck it by hand the same way.

### What the "Not verified here" section could not establish

This run **is** in `krlmlr/duckdb-r` and is readable:

| field | value |
|---|---|
| run | [31115199238](https://github.com/krlmlr/duckdb-r/actions/runs/31115199238) |
| workflow | `rcc` — `.github/workflows/R-CMD-check.yaml` |
| event | `push` |
| `head_branch` | `v1.4-andium-build-base` |
| `head_sha` | `6e7dd75ce` |
| started / concluded | 15:18:29Z / 15:59:46Z, `failure` |

So the workflow that reaches `commit/action.yml` on a series ref is `rcc`,
reached by an ordinary `push` event.

### Why narrowing the trigger is not the fix

`main`'s `R-CMD-check.yaml` fires on
`main`, `master`, `release`, `next`, `cran-*`.
The copy **at `6e7dd75ce`** — the buffer commit the ref points at — has one more:

```yaml
on:
  push:
    branches:
      - main
      - master
      - release
      - next
      - cran-*
      - v*.*-*        # ← matches v1.4-andium-build-base
```

`v*.*-*` matches every `v1.4-andium-*` and `v1.5-variegata-*` ref there is.
It was already dropped on `main`, which is why no `main-*` ref is affected —
but `series-loop.md` is explicit that `-build` and `-build-base`
**carry no ports by design**
("its tooling refreshes only at a forward"),
so those refs keep the old trigger indefinitely,
and no change to `main`'s workflow list can reach them.
The guard in `commit/action.yml` proposed here is what does,
because the action is read from the same frozen tree
and the guard travels with it once the buffer next refreshes.

### The loop is self-reproducing

The push at 15:18 that restored `v1.4-andium-build-base` to `6e7dd75ce`
*is* the push event that produced the auto-commit at 15:52.
The tree at that commit carries `Config/roxygen2/version: 8.0.0.9000`
while CI's rolling roxygen2 is `8.1.0.9000`,
so the drift is guaranteed to be there,
and the hand-fix re-creates the condition it was fixing.
This firing's hand-fix will very likely do the same —
so expect a third occurrence at the next firing,
and one per firing until this merges.

Worth noting alongside #2531, which removes the *drift*
this particular auto-commit is made of.
The two are independent: #2531 stops `Config/roxygen2/version`
being a source of drift at all,
this PR stops any drift being committed onto a series ref.
Either alone would have prevented both occurrences so far;
only this one prevents the next kind.

Per `series-loop.md` §7, that is now two firings' worth of
hand-unsticking paid on this cause.

---

Issue #2522, opened 2026-08-04
Inbox zero


krlmlr:
The full triage — verdicts, evidence, and drafted closing comments for
every open issue and PR — is in
[`plan/PLAN-inbox-zero.md`](https://github.com/krlmlr/duckdb-r/blob/claude/review-open-issues-prs-46alwv/plan/PLAN-inbox-zero.md);
the verdict vocabulary and the no-knowledge-losing-closes rule are stated in
[`handbook/operations/triage/`](https://github.com/duckdb/duckdb-r/blob/main/handbook/operations/triage/README.md).
This issue is the execution index: plain references, no checklist —
each item's own closure is the tracking signal.
A close credits the reporter, names the reopen condition, and,
where no code changed, lands its answer in the owning handbook leaf
in the same PR.

## Review PRs

* #2499 — implements the #455 sketch exactly; review the test-only injection parameter, merge
* #643 — invite a rebase down to the two remaining translations; the bugfix half already landed on main

## Fixes staying open, vehicle named

* #1796 (P1) via #1797
* #455 via #2499
* #2425 via krlmlr/duckdb-r#106
* #184 via #2401
* #97 via #1027 plus an upstream ping
* Pre-located one-liners: #155, #1982, #172, #1052
* #1829 — cosmetic `-Wall` noise from vendored headers; we create patches, fix belongs in duckdb/duckdb

### Scheduled

* #1511 (retires the CSV cluster, then #118's stopgap)
* #126
* #1997 (streaming per #2413's design)
* #2029
* #22 (scope added 2026-08-02)

### 2026-08 upkeep cohort

* #2501
* #2498 

## Hard problems

* #1797 — rebase, restore structured error text; closes #1796, the one P1
* #1027 — answer the contributor's pending question first, then choose land-now vs gate on #2413
* #72 — engine four majors on; the measurable remainder is #97
* #98 — no second registration backend materialized in 2.4 years
* #1065 — whole-result Arrow fetch is inherently RAM-proportional; streaming alternatives exist

## Roadmap epics (deliberately open)

* #2234 — Windows extension toolchain (absorbs #100, #1581, #2503 residue)
* #117 — spatial interop (absorbs #1670)
* #66 — webR · #1585 — blocked on tidyverse/dbplyr#1838 · #118 — ingestion API
* #209 — register-vs-write decision · #2365 — jemalloc before the 2.0 vendor · #540 — relational-API debt · #2497 — dbplyr dialect upstream

## Wait

[...]

PR #2551, opened 2026-08-07
docs(types): State the UTF-8 repair that works, not the one that looks like it

### LINKED ISSUE #12: Error: Invalid Input Error: Invalid unicode (byte sequence mismatch) detected in segment statistics update
hi, i'm still hitting [this error](https://github.com/duckdb/duckdb/issues/1650) on both the CRAN version and also the `duckdb_0.8.1-9000`  dev version..

	# using github
	remotes::install_github("duckdb/duckdb-r")

	# minimal reproducible example
	library(duckdb)
	con <- dbConnect( duckdb::duckdb() , dbdir = 'my-db.duckdb' )
	my_df <- structure(list(no_municipio_esc = "Est\xe2ncia", no_municipio_prova = "Est\xe2ncia"), row.names = 16L, class = "data.frame")
	dbWriteTable( con , 'my_table' , my_df )
	
	
console output:

	> library(duckdb)
	Loading required package: DBI
	> con <- dbConnect( duckdb::duckdb() , dbdir = 'my-db.duckdb' )
	> my_df <- structure(list(no_municipio_esc = "Est\xe2ncia", no_municipio_prova = "Est\xe2ncia"), row.names = 16L, class = "data.frame")
	> dbWriteTable( con , 'my_table' , my_df )
	Error: rapi_execute: Failed to run query
	Error: Invalid Input Error: Invalid unicode (byte sequence mismatch) detected in segment statistics update
	In addition: Warning message:
	Database is garbage-collected, use dbDisconnect(con, shutdown=TRUE) or duckdb::duckdb_shutdown(drv) to avoid this. 
	> 
	> sessionInfo()
	R version 4.3.1 (2023-06-16 ucrt)
	Platform: x86_64-w64-mingw32/x64 (64-bit)
	Running under: Windows 10 x64 (build 19045)

	Matrix products: default

	locale:
	[1] LC_COLLATE=English_United States.utf8  LC_CTYPE=English_United States.utf8    LC_MONETARY=English_United States.utf8 LC_NUMERIC=C                           LC_TIME=English_United States.utf8    

	time zone: America/New_York
	tzcode source: internal

	attached base packages:
	[1] stats     graphics  grDevices utils     datasets  methods   base     

	other attached packages:
	[1] duckdb_0.8.1-9000 DBI_1.1.3        

	loaded via a namespace (and not attached):
	[1] compiler_4.3.1 tools_4.3.1   
	>

krlmlr:

`usage/types/` already said UTF-8 strictness is deliberate, but the recipe it offered — `iconv(x, to = "UTF-8")` or `enc2utf8()` — does nothing for the case the issue is about. Verified on duckdb 1.5.5: for `"Est\xe2ncia"`, `Encoding()` is `"unknown"`, so `enc2utf8()` returns it unchanged and still invalid; `read.csv()` on a latin1 file produces exactly that. Only `iconv(x, from = "latin1", to = "UTF-8")` repairs it, and `iconv()` signals failure as `NA` rather than an error.

So the leaf now names the source encoding, points at `validUTF8()` as the test that agrees with the engine (the console does not — R prints the bytes happily), and says the cheapest fix is at the reader, which is the answer given in the thread.

Thanks to ajdamico for the report and the reprex, and to eli-daniels for the note that already-corrupted data hits this too.

Reopen condition: a string that `validUTF8()` accepts but DuckDB still rejects — that would be an engine bug rather than an encoding one.

Closes #12.

### DIFF
--- a/handbook/usage/types/README.md
+++ b/handbook/usage/types/README.md
@@ -11,7 +11,17 @@ and [`src/transform.cpp`](/src/transform.cpp) (the way back).
   DuckDB checks string validity and rejects invalid UTF-8;
   this is deliberate engine behavior, not a bug
   ([#12](https://github.com/duckdb/duckdb-r/issues/12)).
-  Convert first: `iconv(x, to = "UTF-8")` or `enc2utf8()`.
+  R is the lenient side: it carries the bytes and prints them,
+  so `validUTF8()`, not the console, is what agrees with the engine.
+  Repairing means naming the encoding the bytes are actually in —
+  `iconv(x, from = "latin1", to = "UTF-8")` —
+  and `iconv()` yields `NA` for what it cannot convert
+  unless `sub =` says otherwise,
+  so a repair that fails shows up as missing data.
+  `enc2utf8()` re-encodes only what R has marked, and a string read
+  from a file whose encoding the reader was not told is marked
+  `"unknown"`: it passes through untouched, still invalid.
+  Which is why the cheapest place to fix this is the reader.
 * **Geometry comes back as WKB by default.**
   `dbConnect(geometry = )` chooses: `"blob"`, the default set in
   [`R/dbConnect__duckdb_driver.R`](/R/dbConnect__duckdb_driver.R),


krlmlr:
Apologies for the AI noise.

Issue #179, opened 2024-06-06
Prepared statement with semicolon separated SQL statements


ggrothendieck:
In the example below we have a string containing two semicolon-separated SQL statements.  If the ? for the prepared statement is in the second SQL statement then it works but not if it is in the first.

```
library(duckdb)
con <- dbConnect(duckdb())
dbWriteTable(con, "mtcars", mtcars)

# 1. works (? is in second SQL statement)
rs <- dbSendQuery(con, "update mtcars set hp = 88; select * from mtcars limit ?", params = list(4))
dbFetch(rs)
##    mpg cyl disp hp drat    wt  qsec vs am gear carb
## 1 21.0   6  160 88 3.90 2.620 16.46  4  1    4    4
## 2 21.0   6  160 88 3.90 2.875 17.02  4  1    4    4
## 3 22.8   4  108 88 3.85 2.320 18.61  4  1    4    1
## 4 21.4   6  258 88 3.08 3.215 19.44  3  0    3    1

# 2. gives indicated error (? in first SQL statement)
rs <- dbSendQuery(con, "update mtcars set hp = ?; select * from mtcars limit 2", params = list(4))
## Error: rapi_prepare: Failed to execute statement update mtcars set hp = ?; select * from mtcars limit 2
## Error: Invalid Error: Expected 1 parameters, but none were supplied

dbDisconnect(con)
```

krlmlr:
Thanks. What's your use case, could you call `dbSendQuery()` twice?

ggrothendieck:
Yes. I could do that as a workaround but it would be more convenient if it just worked.  Maybe this requires a fix in duckdb itself?  

Apparently [duckdb 1.0 was just released](https://duckdb.org/2024/06/03/announcing-duckdb-100.html).  Once the duckdb R package includes it I will try it again in case this is basically a duckdb problem.

krlmlr:
Thanks. This behavior is consistent with the intention of the implementation. In a multi-query statement, all queries but the last are executed immediately. Changing this is a major effort.

I could imagine an `sqlSplit()` generic in DBI, and an implementation in duckdb. Or perhaps just a `duckdb_sql_split()` in this package, calling `ExtractStatements()` internally. Would that help?

ggrothendieck:
If it is too difficult I would forget about it since it is not critical but if the splitting ideas are easy (perhaps if duckdb already provides some help with that) then either of those ideas would be useful.

Issue #126, opened 2024-03-24
Deprecate `dbConnect(dbdir, config, read_only)`


krlmlr:
These belong in the `duckdb()` object and must be defined there, their presence in `dbConnect()` and their behavior causes surprises.

PR #2553, opened 2026-08-07
docs(connections): State where the instance's settings bind, and both ways `dbConnect()` gets them wrong

### LINKED ISSUE #83: parameters passed using  dbConnect( config = list()      ) are ignored
Hi,
I saw some code that used config=list() to pass parameters to a duckdb connection.  
I tried it, but for me, they are ignored as illustrated below.
Passing parameters using dbExecute(con, "PRAGMA threads=1; PRAGMA memory_limit='1GB';")  appears to work.

Maybe config= list() is not a thing, just wanted to make sure:

```
library(dplyr)
library(dbplyr)
library(duckdb)

duckdb_path <- "/devroot/sandbox/tmp/duckdb.duckdb"

con <- dbConnect(duckdb::duckdb(dbdir = duckdb_path), config=list("memory_limit"="1GB", "threads" = "1") )
dbGetQuery(con, "select current_setting('memory_limit')")  #                         432.6GB
dbGetQuery(con, "select current_setting('threads')")  # 64
dbExecute(con, "PRAGMA threads=1; PRAGMA memory_limit='1GB';")
dbGetQuery(con, "select current_setting('memory_limit')")    # 1GB
dbGetQuery(con, "select current_setting('threads')")   # 1

```
```

 Session info ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 setting  value
 version  R version 4.0.2 (2020-06-22)
 os       OpenShift Enterprise
 system   x86_64, linux-gnu
 ui       RStudio
 language (EN)
 collate  en_CA.UTF-8
 ctype    en_CA.UTF-8
 tz       America/Toronto
 date     2024-02-23
 rstudio  2023.06.0+421.pro1 Mountain Hydrangea (server)
 pandoc   NA

─ Packages ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 ! package     * version date (UTC) lib source
 P cli           3.6.2   2023-12-11 [?] RSPM (R 4.0.5)
 P DBI         * 1.2.1   2024-01-12 [?] RSPM (R 4.0.5)
 P dbplyr      * 2.4.0   2023-10-26 [?] RSPM (R 4.0.5)
 P dplyr       * 1.1.4   2023-11-17 [?] RSPM (R 4.0.5)
 P duckdb      * 0.9.2-1 2023-11-28 [?] RSPM (R 4.0.5)
 P fansi         1.0.4   2023-01-22 [?] CRAN (R 4.0.2)
 P generics      0.1.3   2022-07-05 [?] RSPM (R 4.0.5)
 P glue          1.6.2   2022-02-24 [?] RSPM (R 4.0.5)
 P lifecycle     1.0.3   2022-10-07 [?] RSPM (R 4.0.5)
   magrittr      2.0.3   2022-03-30 [1] RSPM (R 4.0.5)
 P pillar        1.9.0   2023-03-22 [?] RSPM (R 4.0.5)
 P pkgconfig     2.0.3   2019-09-22 [?] RSPM (R 4.0.3)
   R6            2.5.1   2021-08-19 [1] RSPM (R 4.0.5)
   renv          1.0.3   2023-09-19 [1] RSPM (R 4.0.2)
 P rlang         1.1.3   2024-01-10 [?] RSPM (R 4.0.5)
 P rstudioapi    0.15.0  2023-07-07 [?] RSPM (R 4.0.5)
 P sessioninfo   1.2.2   2021-12-06 [?] RSPM (R 4.0.5)
 P tibble        3.2.1   2023-03-20 [?] RSPM (R 4.0.5)
 P tidyselect    1.2.0   2022-10-10 [?] RSPM (R 4.0.5)
 P utf8          1.2.3   2023-01-31 [?] RSPM (R 4.0.5)
 P vctrs         0.6.5   2023-12-01 [?] RSPM (R 4.0.5)

[1] xxxxxxx /renv/library/R-4.0/x86_64-pc-linux-gnu
 [2] /opt/R/4.0.2/lib/R/library

 P ── Loaded and on-disk path mismatch.
```

### LINKED ISSUE #171: Simplify `dbdir` entrypoint
At present, there are two places where the database directory can be specified for a DuckDB database instance. Either in the **driver** or in the **connection** functions respectively. 

By having two places to specify the dbdir argument it is a bit confusing to know which is the preference. Then, there is a possible point of confusion where a database path may be specified in the driver but another in the connection function. And when this happens the path specified in `dbConnect()` silently takes precedence. 

My _personal_ preference would be that there is only one place to specify the path (it _feels_ best to be in the driver, imo). Otherwise, there should be an error or a warning when setting `dbir` in `dbConnect()` when one is already provided by `duckdb()`

``` r
library(duckdb)

dbConnect(duckdb("mydb.duckdb"))
#> <duckdb_connection 76150 driver=<duckdb_driver dbdir='/reprex-7acf7dd015ed-sappy-tern/mydb.duckdb' read_only=FALSE bigint=numeric>>

dbConnect(duckdb(), "mydb.duckdb")
#> <duckdb_connection 8e300 driver=<duckdb_driver dbdir='/reprex-7acf7dd015ed-sappy-tern/mydb.duckdb' read_only=FALSE bigint=numeric>>

dbConnect(duckdb("a-different.duckdb"), "mydb.duckdb")
#> <duckdb_connection 3c7a0 driver=<duckdb_driver dbdir='/reprex-7acf7dd015ed-sappy-tern/mydb.duckdb' read_only=FALSE bigint=numeric>>
```

<sup>Created on 2024-05-24 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

krlmlr:
Two of the "close with a handbook entry" items in #2522, which groups them on one line — and now folds #171's half in here too, so `usage/connections/` carries one rule for the whole argument family rather than two bullets with opposite shapes. PR 2554 is closed in favour of this one.

`usage/connections/` said `config` binds at instance creation and a reuse ignores it. True, but it does not let a reader predict their own call, and it said nothing at all about the other half: that `dbdir` on `dbConnect()` behaves in the opposite direction.

Verified on duckdb 1.5.5:

- `dbConnect(drv, config = list(memory_limit = "1GB"))` on a reused instance reports 12.5 GiB; the same call with `dbdir =` pointing elsewhere reports 953.6 MiB. That is nbc's three-way comparison in #83's thread, and the reason is that the differing path is what makes the call create an instance.
- `dbConnect(duckdb("a.db"), "b.db")` connects to `b.db`, while the driver object passed in still holds `a.db` and its own live instance — so a reader ends up holding a driver their connection is not on, which is what makes `duckdb_shutdown(drv)` on it look inert.
- `SET memory_limit` on the connection works either way, which is the workaround #83's reporter had already found.

The leaf now names `duckdb()` as where all five belong, states both failure directions, and marks the behaviour provisional against the two issues that end it: #2560 for failing loudly on a collision, #126 for taking the arguments out of `dbConnect()` altogether. Previously the warning was credited to #126 alone.

Thanks to SimonCoulombe and nbc on #83, and JosiahParry on #171 — the reprexes in both threads are what make the precedence visible rather than mysterious.

Reopen condition: #2560 or #126 landing, either of which replaces the documentation with a diagnostic.

Note: touches the same leaf as the PR for #179 (2555), in a different bullet; they should merge in either order.

Closes #83. Closes #171.

### DIFF
--- a/handbook/usage/connections/README.md
+++ b/handbook/usage/connections/README.md
@@ -24,8 +24,15 @@ The load-bearing facts:
   a call that reuses a cached instance ignores them silently
   ([#83](https://github.com/duckdb/duckdb-r/issues/83),
   [#171](https://github.com/duckdb/duckdb-r/issues/171)).
+  `duckdb()` is therefore where they belong.
+  `dbConnect()` takes them too, merged over the driver's,
+  and they land exactly when that call is the one creating the
+  instance — which is why the same `config` argument takes effect
+  beside a `dbdir` naming another file, and is dropped without it.
   To apply new values to a file database,
   release the instance with `duckdb_shutdown()` first.
+  A setting the engine also accepts after startup, `memory_limit`
+  and `threads` among them, can be `SET` on the connection instead.
   Warning in exactly the surprise cases is planned
   ([#126](https://github.com/duckdb/duckdb-r/issues/126)).
 * `dbDisconnect()` closes one connection only;
--- a/handbook/usage/connections/README.md
+++ b/handbook/usage/connections/README.md
@@ -33,8 +33,11 @@ The load-bearing facts:
   release the instance with `duckdb_shutdown()` first.
   A setting the engine also accepts after startup, `memory_limit`
   and `threads` among them, can be `SET` on the connection instead.
-  Warning in exactly the surprise cases is planned
-  ([#126](https://github.com/duckdb/duckdb-r/issues/126)).
+  The silence is provisional in two steps:
+  failing loudly where these arguments collide is
+  [#2560](https://github.com/duckdb/duckdb-r/issues/2560),
+  and taking them out of `dbConnect()` altogether is
+  [#126](https://github.com/duckdb/duckdb-r/issues/126).
 * `dbDisconnect()` closes one connection only;
   its `shutdown` argument is unused.
   Instances are shut down when the driver is garbage-collected
--- a/handbook/usage/connections/README.md
+++ b/handbook/usage/connections/README.md
@@ -19,19 +19,23 @@ The load-bearing facts:
   so reuse is what lets repeated
   `dbConnect(duckdb(dbdir = "my.db"))` calls work at all.
   An in-memory database is never cached.
-* `config`, `read_only`, `home`, and `shared_home`
-  bind when the instance is *created*;
-  a call that reuses a cached instance ignores them silently
+* `dbdir`, `config`, `read_only`, `home`, and `shared_home`
+  all describe the *instance*, so they bind when it is created —
+  and `dbConnect()` accepts every one of them anyway,
+  in two ways, neither of them what a caller wants
   ([#83](https://github.com/duckdb/duckdb-r/issues/83),
   [#171](https://github.com/duckdb/duckdb-r/issues/171)).
-  `duckdb()` is therefore where they belong.
-  `dbConnect()` takes them too, merged over the driver's,
-  and they land exactly when that call is the one creating the
-  instance — which is why the same `config` argument takes effect
-  beside a `dbdir` naming another file, and is dropped without it.
+  `dbdir` wins: it overrides the path the driver was built with,
+  silently, and opens the connection on a driver of its own,
+  leaving the object passed in holding its own database.
+  The rest lose: merged over the driver's, they land only where that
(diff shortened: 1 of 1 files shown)


krlmlr:
Apologies for the AI noise.

Issue #2498, opened 2026-08-02
Support `immediate = TRUE` for direct un-prepared statements


krlmlr:
Even if the C++ code path is the same. For consistency.

PR #2555, opened 2026-08-07
docs(statements): Move the multi-statement prepare rule to the leaf that owns it

### LINKED ISSUE #179: Prepared statement with semicolon separated SQL statements

krlmlr:
One of the "close with a handbook entry" items in #2522, which names `usage/statements/` as the owner.

The fact was already written, but in `usage/connections/`, whose scope is instances, their caching, and shutdown. Where a `?` binds and when the earlier statements of a multi-statement string run is a statement fact. Moving it also gives `statements/` its first stated departure from the DBI baseline — the thing that leaf said it would own — so its "to deepen" note now lists only what is still missing.

Re-verified on duckdb 1.5.5 before moving it: `?` in the last statement binds; `?` in the first fails with "Values were not provided"; and after merely preparing `"insert into probe values (1); select 1"`, `probe` holds a row. Text unchanged apart from one added sentence naming the way out (one statement per call).

Thanks to ggrothendieck for the report — the two-case reprex is what makes the rule visible rather than mysterious.

Reopen condition: #2498, which will make `immediate =` a read argument rather than one silently absorbed by `...`.

Note: touches the same leaf as the PR for #83 and #171 (2553), in a different bullet; they should merge in either order.

Closes #179.

### DIFF
--- a/handbook/usage/connections/README.md
+++ b/handbook/usage/connections/README.md
@@ -32,16 +32,6 @@ The load-bearing facts:
   its `shutdown` argument is unused.
   Instances are shut down when the driver is garbage-collected
   or the session ends.
-* In a multi-statement string,
-  everything before the final statement executes at prepare time,
-  and `?` placeholders bind only in the last statement
-  ([#179](https://github.com/duckdb/duckdb-r/issues/179)).
-  DBI's `immediate = TRUE` is no way around this and no way to opt out:
-  the driver has no unprepared path — every route reaches
-  [`src/statement.cpp`](/src/statement.cpp)'s prepare, which is where the
-  earlier statements run — and the argument lands in `...` unread,
-  which it will stop doing
-  ([#2498](https://github.com/duckdb/duckdb-r/issues/2498)).
 
 *To deepen: absorb the instance and caching section of `?duckdb`;
 drain [#172](https://github.com/duckdb/duckdb-r/issues/172),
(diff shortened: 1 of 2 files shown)


krlmlr:
Apologies for the AI noise.

PR #2557, opened 2026-08-07
docs(integrations): Answer the frame-library request, Arrow and non-Arrow

### LINKED ISSUE #642: Feature request: Copy the Python API's ability to write directly to Polars
It would be great if DuckDB-R could write directly to all the "fast" data formats in R (data.table, collapse, Polars, etc.), but the easiest to implement might be writing to Polars, since the Python API can already do this.

krlmlr:

`usage/integrations/` already named the Arrow stream as the answer to "write directly to Polars", but the issue asks for data.table and collapse too, and those are the opposite case — so they now get their own section rather than a clause inside Arrow, which is the section for libraries that *do* read the stream. The #642 reference appears in both.

**Why no writer**, stated structurally so it does not depend on where those two packages are today: collapse has no container of its own — it computes on R data frames and hands them back, with `qDF()`, `qDT()` and `qTBL()` naming the class to return — and a data.table is those same column vectors under another class. So whatever produces either allocates R vectors, `dbGetQuery()` already does, and a reader either package might grow later would still have R vectors to build. Verified on duckdb 1.5.5 that the relabel after the allocation is genuinely free: `setDT()` and `collapse::qDT()` both leave the column SEXP address unchanged.

**Where the stream still earns its keep**, for a result that should not be held twice:

- `nanoarrow::convert_array_stream(s, to = <prototype>)` builds the target class directly — verified for `data.frame`, `data.table`, and `tbl_df` — rather than a data frame to convert afterwards.
- `dbSendQueryArrow()` with `dbFetchArrowChunk()` converts a batch at a time. The chunk is a `nanoarrow_array`, not a stream, so it is `convert_array()` that takes it.
- One caveat worth stating: a data.table built through `to =` carries no spare column slots (`truelength` 0), so the first `:=` shallow-copies it and warns. `setDT()` on the result fixes that, in place again.

**And the stream is single-use.** The first `convert_array_stream()` over a `dbGetQueryArrow()` result returns the rows, a second over the same object returns zero — it drains rather than erroring, so handing one stream to two consumers yields an empty frame with no complaint.

Thanks to BorgeJorge for the request.

Reopen condition: an R frame library that can consume something faster than the Arrow C stream — that would make a dedicated writer worth its C++.

Closes #642.

### DIFF
--- a/handbook/usage/integrations/README.md
+++ b/handbook/usage/integrations/README.md
@@ -84,13 +84,13 @@ without an R data frame in between —
 `nanoarrow::convert_array_stream()` —
 so a dedicated writer per frame library
 (Polars was the one asked for) is this route, not new C++
-([#642](https://github.com/duckdb/duckdb-r/issues/642)) —
-and a library that does not read the Arrow C stream gains nothing
-from one either, because its input is an R data frame,
-which is what the conversion produces anyway.
-The stream feeds one consumer:
-it drains as it is read, so a second pass over the same object
-sees zero rows rather than the result again.
+([#642](https://github.com/duckdb/duckdb-r/issues/642)).
+A library that cannot read an Arrow stream gains nothing from a
+writer either: its input is an R data frame,
+and that is what converting the stream produces anyway.
+The stream feeds one consumer, draining as it is read,
+so a second pass over the same object sees zero rows
+rather than the result again.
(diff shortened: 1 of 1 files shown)


krlmlr:
Apologies for the AI noise.

PR #2606, opened 2026-08-09
fix(glue): Comment out the names of unused parameters

krlmlr:
One topic out of the compiler-hygiene work for #1829: every `-Wunused-parameter` site in the hand-written glue.

The glue implements engine and R callbacks whose signatures it does not always use in full — replacement scans that ignore the `ClientContext`, ALTREP methods that ignore `deep`/`pvec`/`na_rm`, a cast function that ignores its `CastParameters`. Each name is commented out rather than deleted, so it stays where a reader looks for it while saying nothing to the compiler.

Only definitions change; the declarations in the headers keep their names, so nothing about the documented signatures moves. No behaviour changes.

Files: `src/database.cpp`, `src/register.cpp`, `src/reltoaltrep.cpp`, `src/scan.cpp`, `src/signal.cpp`.

`src/rfuns.cpp` has the same warnings but is vendored from `duckdb-rfuns` by `scripts/vendor-rfuns.sh`, so it is deliberately left out of this PR and handled separately.

How it was found: the glue translation units compiled with `-Wall -Wextra` plus a curated set of extra warnings, with diagnostics attributed to the file they point at, so warnings raised inside the vendored engine headers are not counted against the glue.

### DIFF
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -12,7 +12,7 @@
 
 using namespace duckdb;
 
-static bool CastRstringToVarchar(Vector &source, Vector &result, idx_t count, CastParameters &parameters) {
+static bool CastRstringToVarchar(Vector &source, Vector &result, idx_t count, CastParameters & /* parameters */) {
 	GenericExecutor::ExecuteUnary<PrimitiveType<uintptr_t>, PrimitiveType<string_t>>(
 	    source, result, count,
 	    [&](PrimitiveType<uintptr_t> input) { return StringVector::AddString(result, (const char *)input.val); });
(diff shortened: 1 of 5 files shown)


krlmlr:
Closing: `-Wunused-parameter` is no longer enforced anywhere, so there is nothing here to fix.

The vendored engine cannot be held to it — thousands of hits, nearly all from headers instantiated through templates, and not something a patch stack can carry. The glue implements those same engine callbacks, so holding it to a rule the engine is not held to would only make the two disagree. Same shape as the `-Wredundant-move` decision in #2607.

The category moves to the not-enforced list in #2620, with that reason recorded beside it.

---

PR #2607, opened 2026-08-09
fix(glue): Return without a redundant `std::move()`

krlmlr:
One topic out of the compiler-hygiene work for #1829: the four `-Wredundant-move` sites in the glue.

All four return a local `unique_ptr<Derived>` from a function declared to return `unique_ptr<Base>` — the two replacement scans in `src/register.cpp`, and the global/local table function state initialisers in `src/scan.cpp`. The named return value is already treated as an rvalue there, so the `std::move()` does nothing.

This is P1825, applied as a defect report against C++11, so it is not a C++20-only rule: GCC has implemented it since 9 and Clang since 13, both far below anything that can build a package requiring R >= 4.2. The move is redundant on every compiler this package supports, which is exactly what `-Wredundant-move` reports.

Note that the same warning fires ~630 times inside the vendored engine, where it is upstream's house style rather than a defect; that category is deliberately not enforced against `src/duckdb/`.

---
_Generated by [Claude Code](https://claude.ai/code/session_0148iKGyG7gmmQsdaDckmA3x)_

### DIFF
--- a/src/register.cpp
+++ b/src/register.cpp
@@ -119,7 +119,7 @@ unique_ptr<TableRef> duckdb::EnvironmentScanReplacement(ClientContext &context, 
 	// binds (e.g. during rel_to_altrep()) do not need to look up the data
 	// frame from the environment again.
 	table_function->external_dependency = make_shared_ptr<ExternalDependency>();
-	return std::move(table_function);
+	return table_function;
 }
 
 class RArrowTabularStreamFactory {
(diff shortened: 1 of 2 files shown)


krlmlr:
No performance cost, in either direction — closing in favour of consistency with upstream.

`std::move()` in a return statement only costs something when the returned local has *the same* type as the return type: there it defeats NRVO, turning an elided construction into a move. All four sites here return a `unique_ptr<Derived>` from a function declared `unique_ptr<Base>` — the two replacement scans return `unique_ptr<TableFunctionRef>` as `unique_ptr<TableRef>`, and the two table-function state initialisers return `unique_ptr<DataFrameGlobalState>` / `unique_ptr<DataFrameLocalState>` as their base. Elision cannot apply across those types, so a move happens either way, and the emitted code is the same with or without the `std::move()`.

So the only thing the warning buys is the warning itself, and the engine writes it the other way ~630 times. `-Wredundant-move` is therefore dropped from the enforced set in both scopes rather than fixed in the glue, and the reason is recorded with the warning set.

---

PR #2653, opened 2026-08-13
fix(series-check): Classify an `R CMD check` gate failure instead of giving up on it

krlmlr:
`scripts/series-check.sh` has no classification rule for the shape a red commit most often takes: `R CMD check --as-cran` failing the check gate. Every such failure comes back as `unclassified: review logs2.d/… by hand`.

Today's series-loop firing hit that at full scale. All six series were red end to end — 679 commits in `<S>-green..<S>-dev` — and the script answered `unclassified` for every one, so the firing opened the logs itself to read what each of them said plainly:

```
❯ checking compilation flags used ... WARNING
  Compilation used the following non-portable flag(s):
    '-Wno-redundant-move' '-Wno-unused-parameter'
  including flag(s) suppressing warnings
0 errors ✔ | 1 warning ✖ | 0 notes ✔
Error: R CMD check found WARNINGs
```

(That particular cause is #2652, already merged — this PR is about the diagnosis being unavailable, not about the bug it was hiding.)

The new rule reads the gate's own verdict and quotes the `❯ checking …` headers beside it, which is what says where to look without opening anything:

```
REPAIR c2cc7259c…
       R CMD check found WARNINGs (checking compilation flags used)
```

and, where the injected flags also made the install noisy:

```
       R CMD check found WARNINGs (checking whether package ‘duckdb.dev’ can be installed; checking compilation flags used)
```

Placement in the `classify()` chain:

- behind the snapshot, test and style rules — when one of those matches it names a fix more precisely;
- ahead of the network rule, because `Error: R CMD check found WARNINGs` is a definite answer. `--as-cran` named the checks it is unhappy about, so this is a REPAIR and a rerun buys the same ones again.

`.claude/skills/series-loop.md`'s classification table gains the matching row.

Verified against the two real logs above, taken from `rcc2` at `c2cc7259c` (`main-dev`) and `2a4fbc6b0` (`main-fwd-dev`).

---
_Generated by [Claude Code](https://claude.ai/code/session_018zSTP8ynd8m2TJY1vQEAtN)_

### DIFF
--- a/.claude/skills/series-loop.md
+++ b/.claude/skills/series-loop.md
@@ -401,6 +401,7 @@ is reading; they are the same bytes:
 | `Error ('test-….R:N:M')` | real test failure | fix test/R code at origin |
 | the cause is in `src/duckdb/` itself and the next upstream commit fixes it | upstream was transiently broken | fold the pair into one; if the next one is red too, forward-port (below) |
 | `Changes detected in workflow_dispatch build` | style / roxygen drift | fix formatting at origin |
+| `Error: R CMD check found WARNINGs` / `ERRORs`, with the `❯ checking …` headers above it naming which checks | `R CMD check --as-cran` is unhappy about the tree, or about the environment CI built it in | read the named checks; fix at origin, or on `main` when the cause is the job's own setup |
 | a gate reached out and was refused — `cannot open URL`, `SSL connect error`, a refused or reset connection — while the tests themselves passed | infra, not the tree | rerun the commit: `retry-<S>-dev` (below) |
 | none of the above and no test phase | cancelled or infra | rerun the commit: `retry-<S>-dev` (below) |
 
--- a/scripts/series-check.sh
+++ b/scripts/series-check.sh
@@ -175,6 +175,14 @@ classify() { # <sha> -> "<kind>|<one line>"; kind `transient` means rerun, do no
     echo "test|test failure ($(grep -oE "Error \('test-[^']+'\)" <<<"$log" | head -1))"
   elif grep -q "Changes detected in workflow_dispatch build" <<<"$log"; then
     echo "style|style/roxygen drift"
+  elif grep -qE '^Error: R CMD check found (ERROR|WARNING|NOTE)' <<<"$log"; then
+    # The check gate's own verdict, and it is definite: `--as-cran` named the
+    # checks it is unhappy about, so a rerun buys the same answer. Ahead of the
+    # network rule for that reason, and behind the three above it because a
+    # snapshot, a test or a style diff says more precisely what to fix.
+    echo "check|$(grep -oE 'R CMD check found .*' <<<"$log" | head -1)" \
+      "($(grep -E '^❯ checking' <<<"$log" | sed 's/^❯ //; s/ \.\.\..*//' |
+        tr '\n' '\a' | sed 's/\a$//; s/\a/; /g'))"
   elif grep -qE "$net_re" <<<"$log"; then
     gate=$(failed_gate "$log")
     echo "transient|network failure in the ${gate:-unnamed} gate ($(grep -oE "$net_re" <<<"$log" | head -1))"


krlmlr:
Evidence from the next firing (2026-08-13 17:00 UTC), plus one correction to the placement argument.

`series-check.sh` answered `unclassified` for all three red series again — `main` at `c2cc7259c`, `main-fwd` at `96ce354da`, `v1.5-variegata` at `f863e0984`. Reading every failing log in `<S>-green..<S>-dev` by hand gave a clean split across 166 commits, and nothing but this shape:

- 111 × `checking whether package 'duckdb.dev' can be installed` + `checking compilation flags used`
- 55 × `checking compilation flags used` alone

Distribution: `main` 3, `v1.5-variegata` 6, `main-fwd` 157. So the rule as written covers 100% of what this firing had to diagnose, and both header variants it quotes occur in the wild — the two-header one being the majority, not the corner case.

**The correction is to the RETRY/REPAIR split, not to the rule.** The argument for placing this ahead of the network rule is "`--as-cran` named the checks it is unhappy about, so this is a REPAIR and a rerun buys the same ones again". A rerun would not have. The flags came from the *job*: `before-install` wrote them into `~/.R/Makevars`, and composite actions are resolved from the branch tip that triggered the run, not from the commit under test. Every one of these 166 commits has a tree with no warning gate in it at all — `git show c2cc7259c:.github/workflows/custom/before-install/action.yml` has no `warnings.sh` line — so the commits were never what failed, and none of them was a repair. With #2652 merged and ported onto all six `-dev` tips, dropping the stale records was enough; the rebuilds are running now and `main`'s shard-1 planned exactly the three.

That is a distinguishable case, and cheap to distinguish: the gate names the flags, and whether the commit's own tree asks for them is one `git show`. Suggestion for the message this PR prints — the header quote is already the right place to hang it — is to say which of the two it is, e.g. `R CMD check found WARNINGs (checking compilation flags used — flags not in this commit's tree; stale harness verdict)`. Where the flags do come from the commit, REPAIR stays exactly right.

I have not touched this branch. The drop side of it is #2654, which adds `scripts/rcc-drop.sh` for the operation this firing did by hand; the two are independent and this one is the diagnosis half.

---

krlmlr:
Hit again on the next firing (2026-08-13, later the same day). `scripts/series-check.sh` answered:

```
=== main-fwd: 506 in flight, 173 buffered, green=447b1ac63 dev=06d2573fd
  REPAIR fb76c0416f8c4218b8041009cb8cb8ab426f1dc8
         unclassified: review logs2.d/fb/fb76c0416f8c4218b8041009cb8cb8ab426f1dc8.log on branch rcc2 by hand
```

and the firing opened that log, and eight more like it, by hand — all nine `0 errors | 1 warning | 0 notes` with `❯ checking compilation flags used ... WARNING` as the sole complaint. So the rule this PR adds is still the diagnosis that is missing.

One refinement worth folding in before it lands, because the message the rule prints is what the next firing acts on. For this particular check, `REPAIR` sends the firing to the wrong tool. The flags come from `~/.R/Makevars`, written by a composite action resolved from the **branch tip** that triggered the run, not from the commit under test — none of the nine trees contains the string `redundant-move` at all, and `scripts/warnings.sh` does not exist at any of them. Amending such a commit changes nothing about why it was red; what clears it is dropping the record so the branch's current tooling judges it again (#2654).

That does not argue for `RETRY` either — the retry pair rebuilds from *that commit's* tree, which is a different question. Perhaps `checking compilation flags used` is worth naming as its own outcome, or the message can say "the gate, not the tree" where the only unhappy check is one the commit cannot have caused. The remaining sub-cases (`checking whether package ... can be installed`, and whatever `--as-cran` reports next) genuinely are repairs, so the rule as written is right for them.

---

krlmlr:
Evidence for a second gap in the same `classify()` chain, from today's firing — adding it here rather than opening a second PR, since it is the same function and the same shape of problem.

An **install failure is classified `transient`**, so the verdict comes back RETRY when the right answer is REPAIR:

```
=== main-fwd: 605 in flight, 74 buffered, green=96ce354da dev=03894c699
  RETRY  84370b8a34056ee270bfadf4b50aeef9c8a527b4
         cancelled or infra (no test phase in log)
```

The tree was genuinely broken. `84370b8a3` carried a class declaration with no definition anywhere, and `R CMD INSTALL` failed on the link:

```
** testing if installed package can be loaded from temporary location
Error: package or namespace load failed for 'duckdb.dev' in dyn.load(...):
  .../duckdb.dev.so: undefined symbol: _ZN6duckdb14RCallbackScopeD1Ev
Error: loading failed
```

68 contiguous commits on `main-fwd-dev` were red this way, every one `failed_stages: ["install"]`, all from run [31743600051](https://github.com/krlmlr/duckdb-r/actions/runs/31743600051). Following the RETRY would have spent the series' one rerun and come back with the same failure; the cause is #2657 and the fix was a stage-2 repair.

The reason it lands in `transient` is the catch-all above `unknown`:

```sh
  elif ! grep -q "test_local\|testthat" <<<"$log"; then
    echo "transient|cancelled or infra (no test phase in log)"
```

An install failure has no test phase *by construction* — `rcc-one.sh` prints `install: failure -- skipping the remaining gates` and exits — so absence of a test phase is not evidence of a cancelled leg here. That is the same trap the file already warns about for the budget kill, one stage earlier.

A rule keyed on the marker our own tooling writes would settle it, quoting the first `Error:` line the way the proposed `R CMD check` rule quotes the `❯ checking …` headers:

```sh
  elif grep -q "install: failure -- skipping the remaining gates" <<<"$log"; then
    echo "install|R CMD INSTALL failed (<first Error: line>)"
```

Placement, by the same reasoning as this PR's rule: behind the snapshot, test and style rules, and behind the network rule — an install that failed because a fetch was refused is still a flake — but ahead of the no-test-phase catch-all, which is what is answering wrongly today.

---

krlmlr:
Second firing in a row that paid for this. Today's run hit the same gap on `main-fwd`: 14 red commits at the tip of `main-fwd-green..main-fwd-dev`, and `series-check.sh` answered

```
REPAIR 5e807f4c1019ce22d3e021a3c66ed56bcc6f000f
       unclassified: review logs2.d/5e/5e807f4c1019ce22d3e021a3c66ed56bcc6f000f.log on branch rcc2 by hand
```

so the firing opened the log itself. What it said, in the shape this PR reads:

```
❯ checking whether package ‘duckdb.dev’ can be installed ... [598s/212s] WARNING
  See below...

0 errors ✔ | 1 warning ✖ | 0 notes ✔
Error: R CMD check found WARNINGs
```

With this PR's rule that reads as `R CMD check found WARNINGs (checking whether package ‘duckdb.dev’ can be installed)`, which points straight at the install log — where the actual cause was two `-Wreturn-type` warnings in `src/duckdb/extension/core_functions/scalar/date/date_part.cpp`, from the `DatePartSpecifier` dispatchers upstream added in duckdb/duckdb#24573. Decided by run [31782440469](https://github.com/krlmlr/duckdb-r/actions/runs/31782440469), shard 14, `failed_stages: ["check"]`, exit 1.

Worth noting the install-check variant specifically: the header names the install, not `checking compilation flags used` as in the two logs already quoted above, so a compiler warning in vendored code surfaces under that header. Repaired this firing by folding `patch/0039` into the breaking vendor commit; the diagnosis is what cost the time, not the fix.

---

PR #2654, opened 2026-08-13
feat(rcc): Add `rcc-drop.sh`, for verdicts the job decided rather than the tree

krlmlr:
The verdict store has no way to say *this verdict was never about the commit*. A rerun overturns one by deciding the commit again, and `retry-<S>-dev` asks for exactly one commit to be judged again — but it runs the workflow from **that commit's** tree, which is a different question from "judge it under the tooling the branch carries today".

That question comes up whenever a workflow change decides commits. The composite actions are resolved from the branch tip that triggered the run, not from the commit under test, so a bad one reaches every build in the job.

Today's series-loop firing hit it at scale. #2620 appended `-Wno-redundant-move -Wno-unused-parameter` to `~/.R/Makevars`, where `R CMD check --as-cran` read them back:

```
❯ checking compilation flags used ... WARNING
  Compilation used the following non-portable flag(s):
    '-Wno-redundant-move' '-Wno-unused-parameter'
  including flag(s) suppressing warnings
Error: R CMD check found WARNINGs
```

166 commits across `main` (3), `v1.5-variegata` (6) and `main-fwd` (157) were red for it, every one of them decided by a run on 2026-08-10/11, and every one of their own trees carrying no warning gate at all. #2652 fixed the cause and every `-dev` tip already carried the port, so the trees were fine and only the records were wrong. The firing removed them by hand and the commits are being rebuilt green as I write this — this PR is that operation, so the next firing does not repeat it with a shell loop over `git rm`.

**What it does.** Removing the record is what makes the commit undecided: work selection reads presence and nothing else (`scripts/rcc-decided.sh`), so the next `each-rcc` run on the branch plans it like a commit that had never been built. The log goes with it, because a log whose record is gone is read by nothing and dated by nothing — `rcc-consolidate.sh` drops such orphans anyway.

```sh
scripts/rcc-drop.sh <sha>...          # or --stdin, one per line
scripts/rcc-drop.sh --dry-run <sha>...
```

**How it writes.** Through `rcc-publish.sh`'s `.remove` staging, which already existed and is already unit-tested, so the drop inherits the whole race protocol — fetch the tip, rebuild the index, retry on a lost push — and adds no second way to write to `rcc2`. Full 40-character SHAs only: an abbreviation names no path in the store, so it would drop nothing and report success, which is the one failure mode that looks like a clean run.

**Two behaviours documented rather than fixed.** The commit's `rcc` status stays red until its rebuild writes a fresh one; it is a display surface and nothing decides from it. And `rcc-logs.yaml` must **not** be dispatched behind a drop — it derives records for undecided commits from exactly those stale statuses, and would put back what the drop removed. Where the sweep is what is wanted (a record a cancelled leg never published), there is nothing to drop first.

**When it is the wrong tool**, stated in the header and the skill: a commit its own tree broke is a repair. Dropping its record costs a rebuild to learn the same verdict and erases the log that said why.

Also here: the `store/` handbook page's writer table gains the row, `.claude/skills/series-loop.md`'s *Dropping a result by hand* paragraph now points at the script instead of describing the file surgery, and `scripts/README.md` is regenerated after adding the file to the ownership map in `docs-readme.R`.

Verified against a local bare remote seeded through `rcc-publish.sh`: a drop removes record and log, a second run of the same drop reports `Already published`, the `--stdin` form drops a record that has no log, and an abbreviated SHA is refused before anything is staged.

---
_Generated by [Claude Code](https://claude.ai/code/session_014L1uEjcXAz5tLoBP6xr1sj)_

### DIFF
--- a/.claude/skills/docs-consistency/docs-readme.R
+++ b/.claude/skills/docs-consistency/docs-readme.R
@@ -138,7 +138,7 @@ groups <- list(
     owner = "handbook/operations/ci/per-commit/store",
     globs = c(
       "each-harvest.sh", "rcc-consolidate.sh", "rcc-cutover.sh",
-      "rcc-lib.sh", "rcc-logs.sh", "rcc-publish.sh",
+      "rcc-drop.sh", "rcc-lib.sh", "rcc-logs.sh", "rcc-publish.sh",
       "rcc-run-fields.jq", "rcc-store-test.sh"
     )
   ),
--- a/.claude/skills/series-loop.md
+++ b/.claude/skills/series-loop.md
@@ -1154,16 +1154,45 @@ The store applies the same newest-run-wins rule to its copy,
 so the leg *replaces* a record rather than appending —
 the one case where a decided commit legitimately changes state.
 
-**Dropping a result by hand is a store-side operation**,
-and only concerns a firing that is reading the store:
-remove `runs2.d/<xx>/<sha>.ndjson` and `logs2.d/<xx>/<sha>.log`,
-then dispatch `rcc-logs.yaml` to re-derive both from the fresh status,
-provided the commit is still inside the store's 30-day window.
-The removal alone no longer does anything:
-nothing sweeps on a schedule to notice the gap,
-so a deletion left unaccompanied is a record simply gone.
+**Dropping a result by hand is a store-side operation**:
+
+```sh
+scripts/rcc-drop.sh <sha>...          # or --stdin, one per line
+```
+
+It removes `runs2.d/<xx>/<sha>.ndjson` and `logs2.d/<xx>/<sha>.log`,
+which is what makes the commit undecided —
+work selection reads presence and nothing else (`rcc-decided.sh`) —
+so the next `each-rcc` run on the branch plans it
+like a commit that had never been built.
 A run's own results are immutable;
 there a newer run is the only thing that supersedes an older one.
+
+**Reach for it when the job decided the commit, not the tree.**
+The composite actions come from the branch tip that triggered the run,
+not from the commit under test,
+so a bad workflow turns commits red
+whose own trees have nothing to do with it,
+at the scale of everything the run touched:
+duckdb/duckdb-r#2620 put `-Wno-*` flags in `~/.R/Makevars`,
+where `R CMD check --as-cran` refused them as non-portable,
+and 166 commits across three series were red for it
+until duckdb/duckdb-r#2652 landed and the records were dropped.
+That is not `retry-<S>-dev`'s question:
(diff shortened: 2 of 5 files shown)


krlmlr:
Evidence from the next firing (2026-08-13, later the same day): the hand-drop this PR replaces was under-applied. Nine commits of exactly this class survived on `main-fwd-dev` — `fb76c0416`, `65776a47f`, `4109a6425`, `ae810a089`, `17cef03c9`, `f4be2754e`, `4ae931dc3`, `92e0afb27`, `d9e5e74f2`. All nine were decided by the same run (`31433179932`, 2026-08-10/11), all nine end `0 errors | 1 warning | 0 notes` with `❯ checking compilation flags used ... WARNING` as the sole complaint, and none of their own trees contains the string `redundant-move` at all — `scripts/warnings.sh` does not exist at any of them.

So the shell loop did not merely cost a firing some typing: it silently missed part of its target set, and the leftovers then read as a live `REPAIR` verdict pointing at a commit whose tree was never at fault. That is the same failure mode the header's "an abbreviation names no path in the store, so it would drop nothing and report success" is guarding against, arriving by a different route.

This firing dropped the nine through `scripts/rcc-publish.sh` with a `.remove` staging — the write path this PR wraps — and did not dispatch `rcc-logs.yaml` behind it. Records and logs are gone from `rcc2`, and the commits are replanned by the ordinary rule on the next `each-rcc` run on the branch.

---

krlmlr:
Correction to the comment above: the shell loop did not miss those nine. It dropped all 166, and an `rcc-logs` run put 110 of them back four minutes later (`710ea515e` on `rcc2`), of which nine were still carrying the restored stale failure when the next firing looked. Nothing dispatched that sweep by hand — `rcc-logs.yaml`'s `push` trigger has no `branches` filter, and `paths` is evaluated over the whole pushed range, so the `main-green` and `v1.4-andium-green` fast-forwards this same firing made matched it through #2578. Details and the fix are in #2655.

So this PR is unaffected — the drop was correct and complete, and the write path it wraps is the right one. But the body's "`rcc-logs.yaml` must **not** be dispatched behind a drop" is stated as advice a firing can follow, and it turned out not to be one a firing controls: the ref moves of stages 3 and 5 were dispatching it. With #2655 that advice becomes enforceable again, and it may be worth saying in the header that the hazard is the sweep running at all behind a drop, whoever started it.

This firing's own drop of the nine was restored the same way — `v1.5-variegata-green` advancing 78 commits triggered run `31743264656`, which put back four of them — and has been re-dropped.

---

krlmlr:
Want to decommission the rcc2 branch and all related tooling.

PR #2665, opened 2026-08-17
fix(series): Register the `ours-version` driver before refusing over it

krlmlr:
`scripts/series-advance.sh` and `scripts/series-port.sh` each refuse up front when `merge.ours-version.driver` is absent, telling the operator to run `scripts/setup-git.sh` — and each then registers the driver itself further down: `series-advance.sh` in its replay branch (added by #2659, given `VENDOR_REPO` by #2663), `series-port.sh` under `--apply`.

The guard reaches a fresh clone first, so the registration below it never runs and the refusal stands. Every firing of the series loop starts in a fresh container, which turns the one step both scripts already know how to take into a manual one on every single run.

Evidence from today's firing (2026-08-17), on the first ref motion of the pass:

```
$ scripts/series-advance.sh v1.4-andium
Error: merge driver not registered, run scripts/setup-git.sh
```

The guard is `scripts/series-advance.sh:88` and `scripts/series-port.sh:117`; the registration each script performs anyway is at `series-advance.sh:499` and `series-port.sh:245`. The firing worked around it by hand and finished, per stage 7 of `.claude/skills/series-loop.md`.

## The change

Register first, refuse only if the driver is still missing. The guard keeps its whole point — refuse rather than stop mid-replay on a `DESCRIPTION` conflict indistinguishable from one a human has to resolve — and the two later call sites become the no-ops they now are, so they go.

`scripts/series-forward-build.sh` refuses the same way and is deliberately left alone: it never registers the driver itself, so its guard is coherent, and it is a one-shot script a human runs during a forwarding rather than something a firing meets every time.

## Verification

- `scripts/series-advance-test.sh`: 54 passed, 0 failed.
- Against a clone with `merge.ours-version.driver` unset, both scripts now register it and carry on (`series-port.sh v1.5-variegata` printed its candidate list, `series-advance.sh v1.5-variegata` reported `green unchanged` / `buffer empty`), where before the patch both exited 1 with the message above.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JowKA86WUSFDnZdNLmh3Ac)_

### DIFF
--- a/scripts/series-advance.sh
+++ b/scripts/series-advance.sh
@@ -83,8 +83,16 @@ fi
 # lives in .git/config and cannot be committed, so a fresh clone has the
 # attribute (.gitattributes) and not the driver, and the replay stops on a
 # DESCRIPTION conflict indistinguishable from one a human has to resolve.
-# Refuse up front, as scripts/series-forward-build.sh already does. After the
-# --abort branch, so a stopped replay can always be discarded.
+#
+# Register it, then refuse if it is still missing. This script already knew how
+# to register it -- below, in the replay branch -- and a refusal reached first
+# made every firing run scripts/setup-git.sh by hand, because a firing runs in a
+# fresh clone. Idempotent, and .git/config is shared with the worktree the
+# replay uses. After the --abort branch, so a stopped replay can always be
+# discarded.
+if [ -x "$(dirname "$0")/setup-git.sh" ]; then
+  VENDOR_REPO="$(git rev-parse --show-toplevel)" "$(dirname "$0")/setup-git.sh" >/dev/null
+fi
 git config --get merge.ours-version.driver >/dev/null ||
   { echo "Error: merge driver not registered, run scripts/setup-git.sh" >&2; exit 1; }
 
@@ -493,11 +501,8 @@ else
   # Every buffer commit bumps DESCRIPTION's vendor counter, so a -dev that has
   # taken a fledge bump conflicts on the `Version:` line at the first replayed
   # commit and at every one after it. That line is what the ours-version merge
-  # driver exists for; register it here as series-port.sh does, so only genuine
-  # conflicts reach the judgement above.
-  if [ -x "$(dirname "$0")/setup-git.sh" ]; then
-    VENDOR_REPO="$(git rev-parse --show-toplevel)" "$(dirname "$0")/setup-git.sh" >/dev/null
-  fi
+  # driver exists for, and the gate at the top of this script has registered it
+  # already, so only genuine conflicts reach the judgement above.
 
   # Where the run stops, and how it says so. The worktree is kept: it holds the
   # conflict, and whoever resolves it needs somewhere to do that. Nothing has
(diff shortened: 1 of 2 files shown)


krlmlr:
Paid again on the 2026-08-19 firing, in a fresh container, on the first read of the pass — `scripts/series-port.sh` this time rather than `series-advance.sh`, and for all six series before any ref had moved:

```
$ scripts/series-port.sh main
Error: merge driver not registered, run scripts/setup-git.sh
```

Same for `main-fwd`, `v1.4-andium`, `v1.4-andium-fwd`, `v1.5-variegata`, `v1.5-variegata-fwd`. Running `scripts/setup-git.sh` by hand once cleared it for the whole firing, which then completed normally (six buffers vendored, four `-dev` branches extended, `main-fwd-green` advanced).

That is the second firing in a row spending the workaround, so the cost is now recurring rather than incidental.

---

krlmlr:
Evidence from the 2026-08-19 21:xx UTC series-loop firing: this is still the first thing a firing pays in a fresh clone.

The firing cloned `krlmlr/duckdb-r` from scratch and the very first stage-3 call refused:

```
$ ./scripts/series-advance.sh main
Error: merge driver not registered, run scripts/setup-git.sh
```

Running `scripts/setup-git.sh` by hand and re-issuing the identical command succeeded, so the refusal cost one round trip and nothing else. Six series were served in the firing; the refusal fires once per clone, not once per series, so the cost is bounded — but it is paid by every firing that starts from a fresh checkout, which is the normal case for a scheduled run in an ephemeral container.

Registering the driver where the script already detects it is missing would have made this firing's first command work.

---

krlmlr:
Paid again on the 2026-08-22 firing — fourth consecutive one — and this time with a likely reason why it is still open.

The workaround itself was unchanged: fresh clone of `krlmlr/duckdb-r`, first ref motion of the pass refused.

```
$ scripts/series-advance.sh v1.4-andium
Error: merge driver not registered, run scripts/setup-git.sh
```

`scripts/setup-git.sh` by hand, re-issue, and the firing completed normally.

What is new is that this PR does not look like it is waiting on review. Its `rcc` commit status has been `pending` since 2026-08-17T16:58:53Z and has never been updated, which is what leaves `mergeable_state: blocked`:

```
statuses: [{state: pending, context: rcc, created_at: 2026-08-17T16:58:53Z, updated_at: 2026-08-17T16:58:53Z,
            target_url: .../actions/runs/32048185095}]
```

The run that status points at finished five days ago and finished green — run 32048185095, `status: completed`, `conclusion: success`, `updated_at: 2026-08-17T17:09:29Z`. So the check passed and the status was never written back to `success`.

It is specific to this PR rather than a repository-wide outage: #2671 (`9d0ac3f`) and #2673 (`67257e2`) both carry `rcc: success` from runs on 2026-08-19.

Re-running `rcc` on `a507de8`, or pushing the branch (it is also 4 commits behind `main`, base `41393bfb` vs. `5af1a5d6`), should let the status settle and unblock the merge. Flagging rather than acting on it: this is not a PR opened by this firing.

---

PR #2678, opened 2026-08-27
fix(build): Ship the `MAKEFLAGS` helper in the tarball, and refuse a value that is not `-jN`

krlmlr:
`configure` fills `MAKEFLAGS` in from `scripts/setup-makeflags.R`, but `.Rbuildignore` lists `^scripts$`, so `R CMD build` strips the directory and no package installed from a tarball can read it. `Rscript` reports a file it cannot open on **stdout** and exits 2, so neither the `2>/dev/null` nor the `|| echo ""` keeps the message out of the capture. Every source install from the built package therefore exports

```
MAKEFLAGS=Fatal error: cannot open file 'scripts/setup-makeflags.R': No such file or directory
```

and then compiles serially — the only trace being a puzzling `Setting MAKEFLAGS=Fatal error: ...` line in the log.

## The change

- Move the helper to `tools/`, which `R CMD build` keeps (verified below), so the same path resolves in the repository and in the tarball.
- Accept the helper's output only when it is the `-jN` it is meant to produce. A helper that goes missing again leaves `MAKEFLAGS` unset instead of exporting whatever came back.
- `handbook/build/configuration/` records both, since it owns the `MAKEFLAGS` knob; `scripts/README.md` is regenerated and the helper's `docs-readme.R` group entry drops with the file.

## Where it showed up

Today's series-loop firing read r-universe and found two reds, both `The action has timed out.` while still compiling `src/duckdb`, and both preceded by the `MAKEFLAGS=Fatal error: ...` line:

- `duckdb.dev` / `macos-oldrel-x86_64`, 3786 s, at `main-green` d794a33b8 — https://duckdb.r-universe.dev/api/actions/logs/98339213208
- `duckdb.dev` / `macos-release-x86_64`, 3812 s, at `main-fwd-green` f6f4c6bdc — https://krlmlr.r-universe.dev/api/actions/logs/98339480947

The serial build is what puts those two targets over r-universe's one-hour budget; the other 13 targets fit either way, which is why this stayed invisible until now. No other series is affected today.

## Verification

- The stdout behaviour reproduces locally: `Rscript missing.R` prints `Fatal error: cannot open file ...` on stdout and exits 2, so `$(... 2>/dev/null || echo "")` captures it.
- `R CMD build` on a minimal package with `^scripts$` ignored keeps `tools/` and drops `scripts/`.
[...]

PR #2684, opened 2026-09-09
feat(series): Write a stage-3 finding into the commit stage 5 mints

krlmlr:
An r-universe failure has no per-commit record anywhere and no commit of its own, so the series keeps it in the message of the next `-dev` commit (series-loop.md stage 3). That commit is minted **and pushed** by `series-advance.sh`, which fuses stages 3 and 5, so today the finding can only be written after the fact: an amend, a force-push, and one `each-rcc` run spent on a commit that is about to be re-minted.

## What it costs today

The 2026-09-09 firing paid it twice, once per lineage, because both universes lost a mac x86_64 target to the same `MAKEFLAGS` cause in the same round:

- `main-green` b26b7ed91 — `macos-oldrel-x86_64` timed out at the one-hour budget, https://duckdb.r-universe.dev/api/actions/logs/102527398737
- `main-fwd-green` d6613dffa — `macos-release-x86_64`, same cause, https://krlmlr.r-universe.dev/api/actions/logs/102529635583

Neither had a fix to carry, so both were findings with no repair — exactly the case stage 3 says to record anyway. `series-advance.sh main` and `series-advance.sh main-fwd` had already minted and pushed 997a8c3e2 and 142ffa2f2 by the time the finding was written, so each was amended to 1a6fff279 / 78466b6f7 and force-pushed, discarding the run each push had just triggered.

Earlier firings paid the same thing: the findings on 16fb13cb5, 71211935f and d6613dffa all sit on commits `series-advance.sh` had already pushed.

## What changes

`series-advance.sh <S> --dev-note <file>` appends the file's text to the newest commit of the chunk, before the push.

- A note takes the replay route. The plain ref move pushes the buffer's own commits unchanged, so there is nothing of this stage's making to write the finding on.
- Asking for a note when the chunk minted nothing is an error, rather than dropping the finding silently.
- A missing or empty note file is refused before any ref is read.

The skill's stage 3 gains the invocation next to the sentence that asks for the finding.

## Verification

`scripts/series-advance-test.sh` — 70 passed, 0 failed (was 57). Claim 13 checks that the note lands on the newest minted commit and not on the one below it, that the commit keeps its own vendor subject, that the fifth version component still rises once per vendor commit, and that both refusals leave every ref where it was. Claim 12, already exercised in the body, is added to the header list it was missing from.

https://claude.ai/code/session_01JfXqgLmHRWbysdWyzeofQ4

---
_Generated by [Claude Code](https://claude.ai/code/session_01JfXqgLmHRWbysdWyzeofQ4)_


krlmlr:
Evidence from the 2026-09-10 firing: the workaround was paid again, once, and this time the discarded run is visible side by side with its replacement.

`series-advance.sh main-fwd` minted and pushed 6f9b635ee, which triggered each-rcc run 305 at 07:49:39 UTC (https://github.com/krlmlr/duckdb-r/actions/runs/34451925657). The r-universe read that followed found `duckdb.dev` 1.5.5.9013.1414 at 418fe1b7b red on `macos-release-x86_64` — the same `MAKEFLAGS` cause as before, no fix to carry, so stage 3's "write it anyway" case. Recording it meant amending 6f9b635ee to fe492b570 and force-pushing at 07:51, two minutes into run 305, which started run 306 on the re-minted commit (https://github.com/krlmlr/duckdb-r/actions/runs/34452094875). Run 305 is left building a tip that no longer exists.

Two details this round adds to the case in the description:

- Only one lineage needed a note, and the run was still discarded. The cost is per note, not per red: `main-dev` read all fifteen targets OK in the same pass and its 18d7e819c was left alone.
- The finding itself was the kind only a commit message keeps. It is the first time this cause has failed the *same* green, version and target twice in a row (previous red 3891 s, this one 3830 s), while the two lineages that rebuilt went green — so the boundary-noise reading every earlier note rests on is what it revises. Had it been written after the fact onto a later commit, it would have named a different green than the one it describes.

No change requested here; `--dev-note` as described would have written it before the push and spent nothing.

---

krlmlr:
Evidence from the 2026-09-10 12:00 UTC firing: paid twice again, and one of the two was not an r-universe finding at all.

`series-advance.sh main` minted and pushed a7ce93a12, `series-advance.sh main-fwd` minted and pushed 4385721b1; both were amended and force-pushed (to 8099ac0de and de321c112), discarding the run each push had triggered.

The `main-fwd` note is the case the description already covers — `duckdb.dev` 1.5.5.9013.1415 at 78466b6f7 lost `macos-release-x86_64` to the same `MAKEFLAGS` cause, third consecutive build, no fix to carry.

The `main` note is a different class, and worth naming because it widens what `--dev-note` is for. Upstream's `ca5ce12b7d` rewrote both call sites that `patch/0036-Mark-two-assert-only-bindings-used.patch` marked, so the entry was retired as part of the replay that conflicted on one of them. That entry sits on the `-dev` branches only — ported from `main` (#2617), never mirrored onto a buffer — so no vendor run could retire it and print why; the commit message is the only place the reason exists. Nothing about it is r-universe-shaped, and stage 5's replay is where it arose rather than stage 3.

So the flag reads better as "a note the stage has to write on the commit it just minted", whatever produced it, than as the r-universe escape hatch specifically — which is what the current `--dev-note` name and the proposed skill placement (next to stage 3's sentence) both suggest. No change requested; the mechanism as described would have written both of these before the push and spent nothing.

---

krlmlr:
The 2026-09-10 firing hit the state this PR treats as a misuse guard: a stage-3 finding with no commit to write it on at all.

Every one of the six series was fully caught up when the firing ran — each `-build` at its upstream branch tip (nothing to vendor), `-dev` level with `-build`, and after stage 3 `-green` level with `-dev`. So `series-advance.sh` minted nothing anywhere: `main` and `main-fwd` reported `buffer empty — vendor`, and `v1.4-andium`'s one buffered commit was dropped as empty (#2682).

r-universe still had a finding, the same `MAKEFLAGS` cause as before, one target per lineage:

- `main-green` 18d7e819c, 1.5.99.9003.1403 — `macos-release-x86_64` timed out at the one-hour budget after 3830 s, https://duckdb.r-universe.dev/api/actions/logs/102929078830
- `main-fwd-green` fe492b570, 1.5.5.9013.1416 — `macos-oldrel-x86_64`, same cause, 3767 s, https://krlmlr.r-universe.dev/api/actions/logs/102929033144

Both logs name it outright (`Rscript scripts/setup-makeflags.R` → `MAKEFLAGS` exported as the error string), and both consequences are a stale published binary rather than a broken one — `mac-x86_64` one version behind on each universe.

With `--dev-note`, `series-advance.sh main --dev-note <file>` would have failed here rather than recorded the finding, and a failing stage 3 is worse than the amend it replaces: it stops the advance for a finding that has no repair to carry. Amending the `-dev` tip was not available either — both tips already carry a verified `success` record and are now `-green`, which moves forward only.

So the note has nowhere to go on a caught-up firing, and this one is not a rare shape: with all six series at their upstream tips it is simply what steady state looks like between upstream pushes. Worth deciding what `--dev-note` should do when the chunk is empty — warn and carry the note to the next firing that mints something, rather than refuse — before the flag becomes the only route stage 3 has.

The finding is recorded here for now, since nothing else keeps it.

---

krlmlr:
Evidence from the 2026-09-11 firing: paid twice again, and the note was a *green* read this time.

`series-advance.sh main` minted and pushed ba17e70ee (run 311, https://github.com/krlmlr/duckdb-r/actions/runs/34595667526) and `series-advance.sh main-fwd` minted and pushed 3992a181a (run 312, https://github.com/krlmlr/duckdb-r/actions/runs/34595698639). Both were amended two minutes later to carry the stage-3 note and force-pushed, to 97085f332 and 5c8adf7e9, starting runs 313 and 314. Runs 311 and 312 are still building tips that no longer exist.

What is new is the shape of the note. r-universe read every one of the six lineages all-fifteen-OK — `main-green` 8099ac0de at 1.5.99.9003.1404 and `main-fwd-green` de321c112 at 1.5.5.9013.1417 among them. On the forward lineage that ends three consecutive builds that lost a mac x86_64 target to the one-hour `MAKEFLAGS` budget, and the three notes recorded for them (on the commits for `0b57b1c6a`, `eeb0648e8` and `ca5ce12b7`) read, by the last of them, as having settled that the target no longer cleared the hour by chance. This build revises that back to the boundary-noise reading, and nothing in the series moved to produce it: #2678 is still an open draft.

So a note is not always a finding — sometimes it is the correction of one, and it is worth as much, because the three notes it revises stay readable forever and a later firing mining them would otherwise inherit the wrong conclusion. That reinforces the framing in the second comment above: the flag is "a note the stage has to write on the commit it just minted", whatever produced it, rather than the r-universe-failure escape hatch. A clean read producing a mandatory note is also the case least likely to be anticipated by a name like `--dev-note` documented next to stage 3's failure sentence.

No change requested.

---

PR #2690, opened 2026-09-12
perf(revdepx): Compile the dev binary through ccache

krlmlr:
The `build` job compiles the whole vendored engine, and it sits on the critical path of every `revdep4` run — no shard starts before the binary exists. Run 34677973468 spent 48 minutes there against a 90-minute timeout, and every earlier run paid the same from scratch: the only thing the job persisted was pak's *package* cache, never compiler output.

Between runs the engine moves by one vendor commit at a time, which is the shape a content-addressed compiler cache pays off on. So this installs `ccache` in the base image, mounts a runner directory as `CCACHE_DIR`, and persists it with `actions/cache`.

**The compilers are reached by PATH, not by `~/.R/Makevars`.** Debian's ccache package keeps one symlink per installed compiler under `/usr/lib/ccache`, so prepending that directory routes every `gcc`/`g++` the build resolves, touching neither R's `Makeconf` nor any `Makevars`. Prefixing the compiler in `Makevars` is the alternative, and it is the one that bit us before: R-devel took the first word of `CC = ccache gcc` and ran a bare `ccache`, which died with `invalid option -- 'E'`. `.github/workflows/install/action.yml` carries that story.

**Only the build job puts `/usr/lib/ccache` on its PATH.** A check container inherits the binary from the base image and never invokes it, so no reverse dependency's own compile changes behaviour or timing.

**Verified end to end** rather than assumed: with `/usr/lib/ccache` prepended and no `Makevars` in play, `R CMD INSTALL` of a small C package compiles as a bare `gcc ... -c hello.c -o hello.o` and ccache records one miss; re-running after deleting the object records a direct hit (1/1).

### Costs, stated plainly

`CCACHE_MAXSIZE` is 2G deliberately: a repository gets 10 GB of Actions cache in total and the universe job's pak cache already spends more than two of them, so an uncapped compiler cache would evict the very thing it shares the budget with.

Editing `base-image.sh` rolls the base image tag, by design. That costs one base-image rebuild and — because the universe image reuses its donor only when the donor stands on the same base — one cold universe build on the first run after this lands. Steady state after that.

### Not in scope

The shard checks compile each reverse dependency twice (old half, then new half) with nothing shared between them. ccache is now present in the image for that too, but wiring it into `check-half.sh` is a separate change with its own correctness story, and its payoff is unmeasured.

https://claude.ai/code/session_019uQbmeLGpgaAd5iAULGbDf

---
_Generated by [Claude Code](https://claude.ai/code/session_011111gzMorCmNxr7sHKWR3L)_

### DIFF
--- a/.github/workflows/revdep4.yaml
+++ b/.github/workflows/revdep4.yaml
@@ -412,6 +412,23 @@ jobs:
           restore-keys: |
             revdepx-pak-build-
 
+      # The compiler cache, which is what keeps this job from recompiling the
+      # whole vendored engine every run. Everything in it was produced by the
+      # base image's toolchain, so the key carries that image's recipe hash
+      # and the R version naming it: a change to either is a different
+      # compiler, and ccache would miss on every object regardless. The run id
+      # makes the key unique, because actions/cache never overwrites a key
+      # that already exists and the cache would otherwise be saved once and
+      # never again; restore-keys then takes the newest cache the same
+      # toolchain left behind.
+      - name: Cache the compiler cache
+        uses: actions/cache@v5
+        with:
+          path: ${{ runner.temp }}/ccache
+          key: revdepx-ccache-${{ env.REVDEPX_R_VERSION_INPUT }}-${{ hashFiles('.github/workflows/revdepx/base-image.sh') }}-${{ github.run_id }}
+          restore-keys: |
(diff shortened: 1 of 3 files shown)


krlmlr:
Not sure this is needed.

PR #2691, opened 2026-09-12
perf(revdepx): Compile the revdeps' checks through ccache

krlmlr:
## Summary

The dev binary is built once per run, so caching it only pays *across* runs — that is #2690's half, and it stays there. This is the other case, and it pays *within* a single run: every reverse dependency is compiled **twice in the same shard**, minutes apart, because the two halves run sequentially and share nothing.

## What changes

- `revdep4/queue.sh` keeps one `.ccache` for the whole shard, exports it, and logs its final size.
- `revdepx/check-half.sh` mounts it as the container's `CCACHE_DIR` and puts `/usr/lib/ccache` on the PATH the check runs under.

Shard-wide rather than per package, because two packages that share a `LinkingTo` dependency compile the same headers, and because the halves already guarantee a second compile to hit on.

## The two halves stay comparable

Inside the container both see identical paths — each mounts its own workdir at the same `/revdepx/out` — so the second half hits on everything that does not depend on the package under test. A package with `LinkingTo: duckdb` reads headers that genuinely differ between halves, and ccache misses on exactly those objects, because what it hashes is the preprocessed source. Nothing the comparison depends on is shared across the halves.

## Knobs and cost

`REVDEPX_CCACHE_DIR=` (empty) turns it off. `CCACHE_MAXSIZE` (5G) bounds the shard's own disk, which lives and dies with the runner and is never uploaded, so the Actions cache budget is untouched.

## Depends on #2690

#2690 installs `ccache` in the base image. Without it `/usr/lib/ccache` does not exist, the PATH entry resolves to nothing, and the checks compile exactly as they do today — inert rather than broken, so merge order is not load-bearing.

## Not measured yet

How much this buys is unknown. The per-half durations the run already records are what will say, which is the reason to run it rather than to argue about it.

https://claude.ai/code/session_019uQbmeLGpgaAd5iAULGbDf


krlmlr:
Not sure this is needed.

PR #2692, opened 2026-09-12
fix(connections): Warn rather than fail when a reused instance drops settings

krlmlr:
Since #2641, `duckdb()` refuses to reuse a cached instance when the call passes `read_only`, a `config` entry, or a storage argument that differs from the one the instance was created with. A reverse-dependency run over 353 packages priced that choice: `datacaged` and `Rduckhts` both stop where 1.5.5 ran, each on a `read_only` handed to a database it had opened earlier.

The setting was silently dropped before #2641 and is still dropped now. What the error bought was the telling, and a warning tells just as loudly without ending somebody else's script. So `warn_instance_settings_ignored()` warns — which is what its name said all along — and `duckdb()` returns the instance it found.

The warning is classed `duckdb_instance_settings_ignored`, so a caller who has read it once can silence it deliberately. The class holds on both paths: `R/rlang.R` gains a `warn()` fallback that honours `class` through `warningCondition()`, because a contract that held only where rlang happens to be installed would not be one.

The `dbdir` override that `dbConnect()` refuses is a separate case and stays an error. It displaces the driver's own database rather than dropping a setting, and it broke no reverse dependency.

`handbook/usage/connections/README.md` argued for the error; that section now carries this reasoning instead.

Alternative approach in #2693; these are mutually exclusive, only one should merge.

https://claude.ai/code/session_016xj5YaYaSnxsdxTibchhKn

### DIFF
--- a/R/Driver.R
+++ b/R/Driver.R
@@ -108,7 +108,8 @@ driver_registry <- new.env(parent = emptyenv())
 #'
 #' Because the instance is created once per database file,
 #' `config`, `read_only`, `home`, and `shared_home` take effect only at creation.
-#' A call that reuses an existing instance cannot apply them, and fails rather than dropping them.
+#' A call that reuses an existing instance cannot apply them, and warns rather than dropping them silently.
+#' The warning is classed `duckdb_instance_settings_ignored`, so it can be caught or suppressed on purpose.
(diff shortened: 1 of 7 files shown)


krlmlr:
Not doing this, keeping the failure and accepting downstream breaks.

PR #2693, opened 2026-09-12
fix(connections): Warn instead of failing for the two callers that break

krlmlr:
An alternative to the blanket dial-back: keep the error #2641 introduced, and hand the same words to `datacaged` and `Rduckhts` as a warning. Those are the two packages a reverse-dependency run over 353 packages found stopping where 1.5.5 ran, each on a `read_only` passed to a database it had opened earlier.

`calling_package()` walks the stack for the innermost namespace that is neither ours nor DBI's, the way `find_caller()` already does for `dbSendQuery()`. A call from a script finds nothing and gets the error.

`INSTANCE_SETTINGS_WARN_ONLY` is meant to be deleted rather than extended: an entry goes when that package stops passing settings to a reused instance, and a third package found later is a reason to reconsider erroring at all rather than to add a line.

**Two limits, stated in the code and the handbook rather than discovered later.** The list does nothing for a user's own script, or any package not on it, doing exactly what these two do. And it makes behaviour depend on the caller, so the same line errors from a script and warns from inside `datacaged` — worth remembering when a report does not reproduce.

The user-facing documentation still says the call fails. A carve-out for two named packages is a concession, not a promise to keep.

Alternative approach in #2692; these are mutually exclusive, only one should merge.

https://claude.ai/code/session_016xj5YaYaSnxsdxTibchhKn

### DIFF
--- a/R/Driver.R
+++ b/R/Driver.R
@@ -417,12 +417,55 @@ check_tz <- function(timezone) {
   timezone
 }
 
+# The CRAN packages that call `duckdb()` with instance settings for a database
+# they opened earlier, and so would stop at the error below.
+#
+# A reverse-dependency run over 353 packages found exactly these two. They get
+# the warning instead; every other caller gets the error. The list is meant to
+# be deleted, not extended: each entry goes when that package stops passing
+# settings to a reused instance, and a third package discovered later is a
+# reason to reconsider erroring at all rather than to add a line here.
+#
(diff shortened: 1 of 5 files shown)


krlmlr:
Working around:

- https://github.com/gecomt/datacaged/issues/4
- https://github.com/RGenomicsETL/duckhts/issues/219

PR #2725, opened 2026-09-13
fix(series-port): Skip what `main` merged for its ancestry alone

krlmlr:
`git cherry` answers "`main` has it and the series does not" by patch-id over `mb..main`, which counts every commit `main` can reach — including a lineage merged for its ancestry alone. Such a merge keeps its first parent's tree, so nothing of the lineage it records ever entered the package `main` builds, and porting one applies a tree from another era on top of this one.

#2713 recorded the v1.1.3-2 tag exactly that way ("Ancestry only: merged with -s ours on the branch, so the tree is unchanged"). The firing that ran after it was offered 13 commits dated 2024-12 to 2025-01 for every non-frozen series — `feat: Limit automatic materialization by number of rows or number of cells (#1017)`, `Add cstdint`, `Skip RE2 tests again`, `Snapshot tests` among them. A default `--apply` would have cherry-picked all of them onto a 2026 tree. The firing worked around it by naming an explicit subset of the candidate list, which is a list somebody has to re-derive by hand every time `main` records a tag this way.

### The change

The candidate walk drops what an ancestry-only merge brought in. The test is the merge's own tree and not its subject: `-s ours` is one way to write "ancestry only", a hand-resolved merge that kept our side is another, and both leave the same fact behind. A merge whose tree did move is untouched, because its side commits did reach main's tree and are still the series' to take.

### Verification

`scripts/series-port-test.sh` is new: it builds a synthetic remote and clone offline — no network, no fixtures on disk, the idiom of `series-advance-test.sh` — and checks five things about the candidate list. An ordinary commit is offered; the two commits of an ancestry-only lineage are not; nor is the merge itself; and a real merge's content still is. Against the unfixed script the two exclusion claims fail and the other three pass, so the suite catches this cause and nothing else.

Checked against the live series too, `--list` before and after:

- `v1.4-andium`: 19 candidates dropped, none gained, every one of them from that merge.
- `main`: the 7 it was offered are gone; only the `fledge:` VERSION bumps remain, which are never auto-picked.
- `v1.5-variegata`: the 13 it was offered are gone.

`docs-links.py`: 142 files, 718 internal links, 0 bad. `bash -n` clean. `scripts/README.md` regenerated by `.claude/skills/docs-consistency/docs-readme.R`.

https://claude.ai/code/session_01GUYJPGtYa4VHmuxrBEbAyF

---
_Generated by [Claude Code](https://claude.ai/code/session_01GUYJPGtYa4VHmuxrBEbAyF)_


krlmlr:
Another firing paid this, on 2026-09-13. Every non-frozen series was offered the same 17 commits from the `-s ours` lineage #2713 recorded:

- `main`: 28 candidates, 11 real — 17 from that lineage
- `v1.5-variegata`: 32 candidates, 15 real — the same 17
- `v2.0-cyanoptera`: 65 candidates, 48 real — the same 17
- `main-fwd`: 0 candidates (its seed sits above them)

`v2.0-cyanoptera` is the one that shows the cost of a default `--apply`, because it is a young series with a real backlog and the two lists are not separable by eye. The run reached `9d18f634d test: Sync tests` and stopped with 198 conflict hunks in `tests/testthat/test-rel_api.R`, every one of them the same shape — a 2024 test file landing on a 2026 tree:

```
<<<<<<< HEAD
=======
  experimental <- FALSE
>>>>>>> 9d18f634d (test: Sync tests)
```

The workaround was the one this PR describes: derive the exclusion set by hand as `git rev-list e74338141^2 --not e74338141^1`, subtract it from the candidate list, and pass the remainder as explicit SHAs to `--apply`. That is a list to re-derive per series, per firing, for as long as this is open. With the exclusion applied all four series ported cleanly and their tooling is now byte-identical to `main`.

One observation offered as a separate data point rather than a change to this PR's scope: `v2.0-cyanoptera` also had to drop `b2fb0968d` by hand (#2698), which patches `src/duckdb/third_party/libpg_query/src_backend_parser_gram.cpp` — a file that series' newer engine no longer has. That is the engine-version class the skill already sends to an explicit subset, not the ancestry class this PR fixes, so the two are distinct; it just means the explicit-subset route is currently carrying both.

---

